PR #19, opened 2015-03-27
(Fixes #9) Added remaining base activity types

### LINKED ISSUE #9: Add support for remaining activity types
* HiveActivity
* HiveCopyActivity
* PigActivity

etc.

sethyates:
Adds:
* AdpCopyActivity
* AdpHiveActivity
* AdpHiveCopyActivity
* AdpPigActivity

Adds:
* AdpRdsDatabase
* AdpJdbcDatabase

### DIFF
--- a/src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
+++ b/src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
@@ -5,9 +5,9 @@ import com.krux.hyperion.objects.aws.{AdpCopyActivity, AdpDataNode, AdpRef, AdpE
 
 case class CopyActivity(
   id: String,
+  input: DataNode,
+  output: DataNode,
   runsOn: Ec2Resource,
-  input: Option[DataNode] = None,
-  output: Option[DataNode] = None,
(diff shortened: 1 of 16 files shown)


src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
```diff
@@ -0,0 +1,35 @@
+package com.krux.hyperion.objects
+
+import com.krux.hyperion.HyperionContext
+import com.krux.hyperion.objects.aws.{AdpCopyActivity, AdpDataNode, AdpRef, AdpEc2Resource, AdpActivity}
+
+case class CopyActivity(
+  id: String,
+  runsOn: Ec2Resource,
+  input: Option[DataNode] = None,
```
realstraw: isn't the input and output required? I think we should force people to give these values rather than give a default None value.

sethyates: Yeah they are but it breaks the fluent interface ;-) I'll make non optional

realstraw:
:+1:

Issue #18, opened 2015-03-27
EC2Resource run other activities


realstraw:
Currently we have Ec2Resource have methods like `runJar` to return activity and populate the runsOn field. Add methods to run other activity like `RedshiftCopyActivity` etc.

Also rename
- `runCopyActivity` to `runCopy`

PR #23, opened 2015-03-29
(Fixes #10) support preconditions

### LINKED ISSUE #10: Support preconditions
Particularly on the ShellCommandActivitys that depend on S3DataNodes. Right now, the activity starts off even if the S3DataNode doesn't exist/is empty. We should at least always add a hard-coded precondition for any Input nodes. Preferrably we would also allow specification of custom preconditions.

cf:

* http://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-concepts-preconditions.html
* http://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-object-s3prefixnotempty.html

sethyates:
Also contains a fix for #18.

### DIFF
--- a/src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
+++ b/src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
@@ -7,16 +7,16 @@ import com.krux.hyperion.HyperionContext
  * Launch a Spark cluster
  */
 case class SparkCluster (
-    id: String,
-    taskInstanceCount: Int,
-    coreInstanceCount: Int,
-    instanceType: String,
-    amiVersion: String,
-    sparkVersion: String,
-    terminateAfter: String
-  )(
-    implicit val hc: HyperionContext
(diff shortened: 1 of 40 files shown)


src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
```diff
@@ -28,26 +28,66 @@ case class SparkCluster (
...
-      Some(instanceType),
-      Some(taskInstanceCount.toString),
-      terminateAfter,
-      keyPair
+  def runMapReduce(name: String) = MapReduceActivity(
+    id = name,
+    runsOn = this
+  )
+
+  def runPigScript(id: String) = PigActivity(
```
realstraw: Some of these parameters are called `id`, some are called `name`, maybe rename one of them for consistency

realstraw:
:+1: :sparkles: :sparkles: just one comment about naming consistency.

PR #28, opened 2015-03-31
(Fixes #5) Support parameters

### LINKED ISSUE #5: Support parameters
We could make some pipelines easily updatable if we supported parameters instead of hard-coding values directly into the pipeline activities.

cf: http://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-custom-templates.html

sethyates:


src/main/scala/com/krux/hyperion/objects/Parameter.scala
```diff
@@ -0,0 +1,102 @@
...
+case class StringParameter(
+  id: String,
+  value: String,
+  description: Option[String] = None,
+  allowedValues: Seq[String] = Seq(),
+  encrypted: Boolean = false
+) extends Parameter {
+
+  def serialize = AdpParameter(
+    name,
```
realstraw: Use the key value initialisation for consistency

realstraw:
:+1:

PR #31, opened 2015-04-01
(fix #14) individual pipeline based cli

### LINKED ISSUE #14: DataPipelineDefs should be executable and we should get rid of the hyperion exectuable
Rather than shipping an executable jar, we should make the DataPipelineDef objects executable.

realstraw:


src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
```diff
@@ -0,0 +1,108 @@
...
+
+  def createPipeline(force: Boolean = false): Option[String] = {
+
+    println(s"Creating pipeline $pipelineName")
+
+    getPipelineId match {
+
+      case Some(pipelineId) =>
+        println("Pipeline already exists")
+        force match {
```
sethyates: why not just good ole
```scala
if (force) { ... } else { ...}
```
?

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
```diff
@@ -2,14 +2,16 @@ package com.krux.hyperion.examples
...
 import com.krux.hyperion.HyperionContext
 import com.typesafe.config.ConfigFactory
 
 /**
  * An example redshift loader object
  */
-class ExampleRedshiftLoad extends DataPipelineDef {
+object ExampleRedshiftLoad extends DataPipelineDef with HyperionCli {
+
+  def pipelineDef = this
```
sethyates: ick.  is there no way around this cruft?

src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala
```diff
@@ -77,3 +77,7 @@ class ExampleSpark extends DataPipelineDef {
   }
 
 }
+
+object ExampleSpark extends HyperionCli {
+  def pipelineDef = new ExampleSpark
```
sethyates: It would be much more pleasant if we could just do something like:
```scala
object ExampleSpark extends DataPipelineDef {
...
}
```
without having to have the `pipelineDef`.

src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
realstraw: I don't have a very good reason. Just think this looks some what clearer. `if` `else` is slightly faster in execution time though. Happy to change it to `if` `else` if that's the preferred way.

src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala
realstraw: The current implementation also supports `object ExampleSpark extends DataPipelineDef with HyperionCli` as shown in the `ExampleRedshiftLoad.scala`

I separate out the `HyperionCli` from `DataPipelineDef`, because I thought maybe there are some use cases that people do not want to mixin the CLI?

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
realstraw: Well, if we have
```scala
trait DataPipelineDef extends HyperionCli
```
then we don't need this, but every pipeline is forced to have a CLI

src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala
sethyates: What is the harm in mixing in the CLI?

src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
sethyates: It made me thing "what's going on here", which is probably a reason to use if/else.

src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala
realstraw: No real harm in any of our use cases, just thought someone may not want to mixin the CLI and decide to build their own uploading and creating logics?

realstraw: Probably I'll change it to have the CLI mixin in DataPipelineDef, and separate them later when we see a use case?

sethyates: Works for me!

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
```diff
@@ -2,14 +2,14 @@ package com.krux.hyperion.examples
 
 import com.krux.hyperion.objects._
 import com.krux.hyperion.Implicits._
-import com.krux.hyperion.DataPipelineDef
+import com.krux.hyperion.{DataPipelineDef, HyperionCli}
```
sethyates: Change not needed anymore?

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
```diff
@@ -2,14 +2,14 @@ package com.krux.hyperion.examples
...
 import com.krux.hyperion.HyperionContext
 import com.typesafe.config.ConfigFactory
 
 /**
  * An example redshift loader object
  */
-class ExampleRedshiftLoad extends DataPipelineDef {
+object ExampleRedshiftLoad extends DataPipelineDef with HyperionCli {
```
sethyates: Change not needed anymore? (apart from object)

src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala
```diff
@@ -1,6 +1,6 @@
 package com.krux.hyperion.examples
 
-import com.krux.hyperion.DataPipelineDef
+import com.krux.hyperion.{DataPipelineDef, HyperionCli}
```
sethyates: Change not needed anymore?

src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala
```diff
@@ -9,7 +9,7 @@ import com.krux.hyperion.expressions.ExpressionDSL._
 import com.krux.hyperion.HyperionContext
 import com.typesafe.config.ConfigFactory
 
-class ExampleSpark extends DataPipelineDef {
+object ExampleSpark extends DataPipelineDef {
```
sethyates: Change not needed anymore?

realstraw: This one is good, there others are removed in new commits.

sethyates:
LGTM :+1: :sparkles2: :sparkles2: :sparkles2: :sparkles2: :sparkles2: :sparkles2:

realstraw:
merged in commit 5249d84

PR #32, opened 2015-04-01
(Fixes #6) Support other aspects of schedules

### LINKED ISSUE #6: Support occurrences and other aspects of schedules
http://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-object-schedule.html

sethyates:


src/main/scala/com/krux/hyperion/objects/Schedule.scala
```diff
@@ -3,26 +3,23 @@ package com.krux.hyperion.objects
...
-    scheduleType: ScheduleType = Cron
-  ) extends PipelineObject {
-
-  val `type` = "Schedule"
+  id: String = "PipelineSchedule",
+  // if None, will use first activation datetime
+  start: Option[DateTime] = None,
+  period: DpPeriod = 1.day,
+  occurrences: Option[Int] = None,
+  end: Option[DateTime] = None,
```
realstraw: IIRC aws datapipeline only allows one of `occurrences` or `endDate`. Not sure if it's over engineer but should we change this to `end: Option[Either[Int, DateTime]]` and remove `occurrences` and change the def to something like this?
```scala
def until(dt: DateTime) = this.copy(end = Some(Right(dt)))
def stopAfter(occurrences: Int) = this.copy(end = Some(Left(occurrences)))
```

sethyates: Yes, I was hoping to rely on the best intentions of the user, but changing as you outline make sense.  I'll make the change.

realstraw:
:+1: :sparkles: :sparkles:

PR #34, opened 2015-04-02
(Fixes #33) Support tags

### LINKED ISSUE #33: Allow to specify tags in PipelineDef and on command line
Tags on command-line are additive.

sethyates:


src/main/scala/com/krux/hyperion/DataPipelineDef.scala
```diff
@@ -30,13 +30,15 @@ trait DataPipelineDef extends HyperionCli {
 
   def defaultObject = DefaultObject(schedule)
 
+  def tags: Map[String, Option[String]] = Map()
+
   def parameters: Iterable[Parameter] = Seq()
 
   def objects: Iterable[PipelineObject] = workflow
-    .foldLeft(Map[String, PipelineObject]())(flattenPipelineObjects)
+    .foldLeft(MutableMap[String, PipelineObject]())(flattenPipelineObjects)
```
realstraw: Why are we changing these to `MutableMap`?

src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
```diff
@@ -24,7 +24,7 @@ class HyperionAwsClient(pipelineDef: DataPipelineDef, customName: Option[String]
     }
   }
 
-  def createPipeline(force: Boolean = false): Option[String] = {
+  def createPipeline(force: Boolean = false, tags: Map[String, Option[String]]): Option[String] = {
```
realstraw: Since we have a default value for `force` maybe have a default value for `tags` as well?

src/main/scala/com/krux/hyperion/DataPipelineDef.scala
```diff
@@ -30,13 +30,15 @@ trait DataPipelineDef extends HyperionCli {
 
   def defaultObject = DefaultObject(schedule)
 
+  def tags: Map[String, Option[String]] = Map()
```
realstraw: Why are we using `Map[String, Option[String]]` instead of just `Map[String, String]`?

sethyates: The value isnt required

src/main/scala/com/krux/hyperion/DataPipelineDef.scala
sethyates: Note the import change above was always Mutable. Renamed so it is (1) more obvious and (2) doesn't clash with the immutable Map. Not sure why it is mutable apart from the way we are folding(?).

src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
sethyates: True. Missed that.

src/main/scala/com/krux/hyperion/DataPipelineDef.scala
sethyates: However having actually finished the implementation it is now obvious that I still have to pass an empty string to withValue or the API fails. Also, the API felt more correct doing Map("tag" -> None) than Map("tag" -> "") for the case where the tag doesn't have a value.

src/main/scala/com/krux/hyperion/DataPipelineDef.scala
realstraw: Ah, right, we were using mutable map to begin with. I think we should change to use the immutable one.

sethyates: I agree. From memory we had to do it as mutable for a reason... But that is the extent of the memory.

realstraw: Interesting, when I change it to immutable map the Spark test fails. Maybe we can leave this one as is, and make it an issue that we could revisit later.

src/main/scala/com/krux/hyperion/DataPipelineDef.scala
realstraw: :+1: then Map("tag" -> None) it is

Issue #35, opened 2015-04-02
Use immutable Map/MapBuilder in DataPipelineDef


realstraw:
Currently we are using MutableMap to concat the Objects ID Map, Ideally it would be better to use immutable Map.

realstraw:
Turns out it's the problem with test case not the mutable Map

realstraw:
Close with 4632448

PR #36, opened 2015-04-05
(fixes #17 #35) make pipeline id transparent and use immutable for objects building

### LINKED ISSUE #17: PipelineObject's id should be transparent
I think user should not worry about ids in general, we currently generate random ids for a few objects like S3DataNode. I think this behaviour should be made global to all other objects, and have a method `withName` or `withDescription` to prepend to the random id for human readability.

realstraw:


src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
```diff
@@ -20,20 +19,23 @@ import com.krux.hyperion.util.PipelineId
...
+  preconditions: Seq[Precondition],
+  onFailAlarms: Seq[SnsAlarm],
+  onSuccessAlarms: Seq[SnsAlarm],
+  onLateActionAlarms: Seq[SnsAlarm]
 )(
   implicit val hc: HyperionContext
 ) extends PipelineActivity {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  @deprecated("use 'withName' instead of 'forClient'", "2015-04-04")
```
sethyates: well... no, this isn't what forClient was for.  We were using `id` to identify what the step was, then if there were multiple clients in the same pipeline then we were using `client` to identify that.  How will that work now?

src/main/scala/com/krux/hyperion/objects/DeleteS3PathActivity.scala
```diff
@@ -11,22 +11,25 @@ import com.krux.hyperion.objects.aws.AdpSnsAlarm
...
+  onFailAlarms: Seq[SnsAlarm],
+  onSuccessAlarms: Seq[SnsAlarm],
+  onLateActionAlarms: Seq[SnsAlarm]
 )(
   implicit val hc: HyperionContext
 ) extends PipelineActivity {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  @deprecated("use 'withName' instead of 'forClient'", "2015-04-04")
+  def forClient(client: String) = this.copy(new UniquePipelineId(client))
```
sethyates: `id =`

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
```diff
@@ -22,7 +21,7 @@ case class Ec2Resource(
   implicit val hc: HyperionContext
 ) extends ResourceObject {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  def forClient(client: String) = this.copy(id = new UniquePipelineId(client))
```
sethyates: Not deprecated like the others.

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
```diff
@@ -33,33 +32,16 @@ case class Ec2Resource(
   def withSecurityGroupIds(securityGroupIds: String*) = this.copy(securityGroupIds = securityGroupIds)
   def withPublicIp() = this.copy(associatePublicIpAddress = true)
 
-  def runJar(name: String) =
-    JarActivity(
-      id = name,
-      runsOn = this
-    )
+  def runJar = JarActivity(this)
```
sethyates: `runsOn = ` for consistency?

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
```diff
@@ -33,33 +32,16 @@ case class Ec2Resource(
...
-      runsOn = this
-    )
+  def runShell = ShellCommandActivity(runsOn = this)
 
-  def downloadFromGoogleStorage(name: String) =
-    GoogleStorageDownloadActivity(
-      id = name,
-      runsOn = this
-    )
+  def downloadFromGoogleStorage = GoogleStorageDownloadActivity(this)
```
sethyates: `runsOn = ` for consistency?

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
```diff
@@ -33,33 +32,16 @@ case class Ec2Resource(
...
 
-  def uploadToGoogleStorage(name: String) =
-    GoogleStorageUploadActivity(
-      id = name,
-      runsOn = this
-    )
+  def uploadToGoogleStorage = GoogleStorageUploadActivity(this)
```
sethyates: `runsOn = ` for consistency?

src/main/scala/com/krux/hyperion/objects/HiveCopyActivity.scala
```diff
@@ -4,23 +4,26 @@ import com.krux.hyperion.HyperionContext
...
+  onFailAlarms: Seq[SnsAlarm],
+  onSuccessAlarms: Seq[SnsAlarm],
+  onLateActionAlarms: Seq[SnsAlarm]
 )(
   implicit val hc: HyperionContext
 ) extends PipelineActivity {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  @deprecated("use 'withName' instead of 'forClient'", "2015-04-04")
+  def forClient(client: String) = this.copy(new UniquePipelineId(client))
```
sethyates: Missing `id = `

src/main/scala/com/krux/hyperion/objects/MapReduceCluster.scala
```diff
@@ -27,49 +27,10 @@ case class MapReduceCluster(
 
   val terminateAfter = hc.emrTerminateAfter
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  def forClient(client: String) = this.copy(id = new UniquePipelineId(client))
```
sethyates: Not deprecated

src/main/scala/com/krux/hyperion/objects/PipelineObject.scala
```diff
@@ -3,6 +3,7 @@ package com.krux.hyperion.objects
 import com.krux.hyperion.objects.aws.{AdpDataPipelineAbstractObject, AdpJsonSerializer,
   AdpPipelineSerializer}
 import com.amazonaws.services.datapipeline.model.{PipelineObject => AwsPipelineObject}
+import scala.language.implicitConversions
```
sethyates: is this required?

src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
realstraw: Since we don't have to worry about conflict ID anymore (which is part of the reason for `forClient`). I'm thinking now have `withName` now fulfil both what the step was and for which client, like `sparkActivity.withName(s"filter_$clientName")`.

src/main/scala/com/krux/hyperion/objects/S3DataNode.scala
```diff
@@ -38,7 +37,7 @@ case class S3File(
   onFailAlarms: Seq[SnsAlarm] = Seq()
 ) extends S3DataNode {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  def forClient(client: String) = this.copy(id = new UniquePipelineId(client))
```
sethyates: Not deprecated?

src/main/scala/com/krux/hyperion/objects/S3DataNode.scala
```diff
@@ -82,7 +81,7 @@ case class S3Folder(
...
```
sethyates: Not deprecated?

src/main/scala/com/krux/hyperion/objects/UniquePipelineId.scala
```diff
@@ -0,0 +1,21 @@
...
+
+class UniquePipelineId(prefix: String) {
+
+  val uniqueId = prefix + "_" + UUID.randomUUID.toString
+
+  override def toString = uniqueId
+
+}
+
+class FixedPipelineId(fixedId: String) extends UniquePipelineId(fixedId) {
```
sethyates: Wouldn't it be better to reverse this and have `FixedPipelineId` as the root with `UniquePipelineId` inheriting it?

```scala
class FixedPipelineId(val uniqueId: String) {
    override def toString = uniqueId
}

class UniquePipelineId(prefix: String) extends FixedPipelineId(prefix + "_" + UUID.randomUUID.toString)
```

src/main/scala/com/krux/hyperion/objects/SparkActivity.scala
```diff
@@ -7,18 +7,19 @@ import com.krux.hyperion.objects.aws.AdpSnsAlarm
...
+  steps: Seq[SparkStep],
+  dependsOn: Seq[PipelineActivity],
+  preconditions: Seq[Precondition],
+  onFailAlarms: Seq[SnsAlarm],
+  onSuccessAlarms: Seq[SnsAlarm],
+  onLateActionAlarms: Seq[SnsAlarm]
 ) extends EmrActivity {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  def forClient(client: String) = this.copy(id = new UniquePipelineId(client))
```
sethyates: Not deprecated.

src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
```diff
@@ -25,50 +25,10 @@ case class SparkCluster (
   val bootstrapAction = s"s3://support.elasticmapreduce/spark/install-spark,-v,$sparkVersion,-x" ::
       hc.emrEnvironmentUri.map(env => s"${hc.scriptUri}deploy-hyperion-emr-env.sh,$env").toList
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  def forClient(client: String) = this.copy(id = new UniquePipelineId(client))
```
sethyates: No deprecated?

src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
```diff
@@ -25,50 +25,10 @@ case class SparkCluster (
...
-      output: Option[DataNode] = None) =
-    HiveCopyActivity(
-      id = id,
-      runsOn = this,
-      filterSql = filterSql,
-      generatedScriptsPath = generatedScriptsPath,
-      input = input,
-      output = output
-    )
+  def runSpark = SparkActivity(this)
```
sethyates: `runOn = `?

src/main/scala/com/krux/hyperion/objects/UniquePipelineId.scala
sethyates: Or just `PipelineId` instead of `FixedPipelineId`?

src/main/scala/com/krux/hyperion/objects/DeleteS3PathActivity.scala
realstraw: Interesting, it still compiles. I'll add it.

src/main/scala/com/krux/hyperion/objects/UniquePipelineId.scala
realstraw: Make sense, I'll change it to `PipelineId` and reverse it.

src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
sethyates: I see (but less fluent) :-)

src/main/scala/com/krux/hyperion/objects/SnsAlarm.scala
```diff
@@ -19,12 +19,23 @@ case class SnsAlarm(
   def withRole(role: String) = this.copy(role = Some(role))
 
   def serialize = new AdpSnsAlarm(
-    id,
```
sethyates: Need to add a withName...

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
sethyates: Also need a withName.

src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
sethyates: Also need a withName.

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
sethyates: Actually, I would get rid of these runners now...  Reason for having them is moot now.

src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
sethyates: Actually, I would get rid of these runners now...  Reason for having them is moot now.

src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
sethyates: OK I've updated my test project and I do think we lose a bit too much expressiveness by losing `forClient`.  Any other solutions than an interpolated string (apart from the obvious non-deprecation)?

src/main/scala/com/krux/hyperion/objects/Ec2Resource.scala
sethyates: Actually, I would get rid of these runners now...  Reason for having them is moot now.

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
```diff
@@ -28,25 +28,24 @@ object ExampleRedshiftLoad extends DataPipelineDef {
...
 
-    Some(RedshiftCopyActivity(
-      id = "copy",
-      input = S3DataNode.fromPath("s3://testing/testtab/").withDataFormat(s3Format),
-      insertMode = RedshiftCopyActivity.OverwriteExisting,
-      runsOn = ec2instance,
-      output = redshiftTable
-    ))
+    Some(
+      ec2Instance.copyIntoRedshift(
```
sethyates: I think this should be:
````scala
Some(
  RedshiftCopyActivity(ec2instance)
    .withInput(S3DataNode.fromPath("s3://testing/testtab/").withDataFormat(s3Format))
    .withOutput(redshiftTable)
    .withInsertMode(RedshiftCopyActivity.OverwriteExisting)
)
````

File: src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala [hunk 18]
```diff
@@ -51,7 +51,8 @@ object ExampleSpark extends DataPipelineDef {
         format(SparkActivity.scheduledStartTime - 3.days, "yyyy-MM-dd")
       )
 
-    val filterActivity = SparkActivity("filterActivity", sparkCluster)
+    val filterActivity = sparkCluster.runSpark
```
sethyates: Would prefer:
````scala
SparkActivity(sparkCluster)
````
over
````scala
sparkCluster.runSpark
````

File: src/main/scala/com/krux/hyperion/objects/SparkCluster.scala (hunk 15 above)
Comment (realstraw) [inline src/main/scala/com/krux/hyperion/objects/SparkCluster.scala]: OK, agreed

File: src/main/scala/com/krux/hyperion/objects/SnsAlarm.scala (hunk 16 above)
Comment (realstraw) [inline src/main/scala/com/krux/hyperion/objects/SnsAlarm.scala]: I did not add `withName` for non-activity objects, since they are not really get displayed on the AWS DataPipeline UI anyway, so maybe the random generated name is good enough.

File: src/main/scala/com/krux/hyperion/objects/PipelineObject.scala (hunk 9 above)
Comment (realstraw) [inline src/main/scala/com/krux/hyperion/objects/PipelineObject.scala]: Yes, we have a implicit conversion of PipelineId to String

File: src/main/scala/com/krux/hyperion/objects/SnsAlarm.scala (hunk 16 above)
Comment (sethyates) [inline src/main/scala/com/krux/hyperion/objects/SnsAlarm.scala]: Certainly do get displayed in UI.  For example, in the Dependencies tab:

```
Input data node    S3DataNode_3cf1a59d-1997-4763-a126-a4ca19cdcf86
Depends on          exportActivity_55a706c4-1321-4c4c-8f5e-d12afa1083a0
Runs on resource Ec2Resource_55a706c4-1321-4c4c-8f5e-d12afa1083a0
```

File: src/main/scala/com/krux/hyperion/objects/DeleteS3PathActivity.scala [hunk 19]
```diff
@@ -11,22 +11,34 @@ import com.krux.hyperion.objects.aws.AdpSnsAlarm
...
+  onFailAlarms: Seq[SnsAlarm],
+  onSuccessAlarms: Seq[SnsAlarm],
+  onLateActionAlarms: Seq[SnsAlarm]
 )(
   implicit val hc: HyperionContext
 ) extends PipelineActivity {
 
-  def forClient(client: String) = this.copy(id = s"${id}_${client}")
+  def withName(name: String) = this.copy(
+    id = id match {
```
sethyates: You should DRY this all up in PipelineId object.

src/main/scala/com/krux/hyperion/objects/DeleteS3PathActivity.scala
```diff
@@ -11,22 +11,34 @@ import com.krux.hyperion.objects.aws.AdpSnsAlarm
...
+      case NameClientObjectId(_, c) => NameClientObjectId(name, c)
+      case _ => NameClientObjectId(name, "")
+    }
+  )
+
+  def forClient(client: String) = this.copy(
+    id = id match {
```
sethyates: Same here you should DRY this all up in PipelineId object.

src/main/scala/com/krux/hyperion/objects/EmrCluster.scala
```diff
@@ -1,3 +1,13 @@
 package com.krux.hyperion.objects
 
-trait EmrCluster extends ResourceObject
+trait EmrCluster extends ResourceObject {
+
+  def runHiveScript = HiveActivity(runsOn = this)
```
sethyates: Remove

src/main/scala/com/krux/hyperion/objects/S3DataNode.scala
```diff
@@ -20,9 +19,9 @@ object S3DataNode {
 
   def fromPath(s3Path: String): S3DataNode =
     if (s3Path.endsWith("/"))
-      S3Folder(PipelineId.generateNewId("S3DataNode"), s3Path, None)
```
sethyates: Well these should be moved to `apply` on `S3Folder` and `S3File`, right?

src/main/scala/com/krux/hyperion/objects/Schedule.scala
```diff
@@ -12,7 +12,7 @@ import com.krux.hyperion.objects.ScheduleType._
  * @note If start time given is a past time, data pipeline will perform back fill from the start.
  */
 case class Schedule(
-  id: String = "PipelineSchedule",
+  id: PipelineObjectId = ScheduleObjectId,
```
sethyates: Why is this not just `PipelineObjectId.fixed("PipelineSchedule")`?  What's the point of having this defined in another place and only used here?

src/main/scala/com/krux/hyperion/objects/Terminate.scala
```diff
@@ -3,12 +3,10 @@ package com.krux.hyperion.objects
 import com.krux.hyperion.HyperionContext
 import com.krux.hyperion.objects.aws.AdpTerminate
 
-class Terminate(
-  implicit val hc: HyperionContext
-) extends PipelineObject {
+object Terminate extends PipelineObject {
 
-  val id = "TerminateAction"
+  val id = TerminateObjectId
```
sethyates: Why is this not just `PipelineObjectId.fixed("TerminateAction")`?  What's the point of having it defined somewhere else and only used here?

src/main/scala/com/krux/hyperion/objects/DefaultObject.scala
```diff
@@ -4,18 +4,21 @@ import com.krux.hyperion.objects.aws.{AdpJsonSerializer, AdpDataPipelineDefaultO
...
 
-  val fields = Map[String, Either[String, AdpRef[AdpDataPipelineObject]]](
-    "scheduleType" -> Left(schedule.scheduleType.toString),
-    "failureAndRerunMode" -> Left(hc.failureRerunMode),
-    "pipelineLogUri" -> Left(hc.logUri),
-    "role" -> Left(hc.role),
-    "resourceRole" -> Left(hc.resourceRole),
-    "schedule" -> Right(AdpRef[AdpSchedule](schedule.id))
-  )
+  val id = DefaultObjectId
```
sethyates: Why is this not just `PipelineObjectId.fixed("Default")`?  What's the point of having it defined somewhere else and only used here?

src/main/scala/com/krux/hyperion/objects/PipelineObjectId.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  override def toString = uniqueId
+
+}
+
+case class FixedObjectId(seed: String) extends PipelineObjectId {
+  override def toString = seed
+}
+
+object ScheduleObjectId extends PipelineObjectId { override def toString = "PipelineSchedule" }
```
sethyates: See the comments on these at their usage sites.  I don't see the point of having these here when they're only used in a single place.

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
realstraw: Cannot really use withInput, withOutput etc because input, output etc are required, otherwise we need to compromise to leave default values.

src/main/scala/com/krux/hyperion/objects/DeleteS3PathActivity.scala
realstraw: will do

src/main/scala/com/krux/hyperion/objects/S3DataNode.scala
realstraw: Yes, will change them

src/main/scala/com/krux/hyperion/objects/Schedule.scala
realstraw: I was thinking the ScheduleObjectId etc. are for internal Hyperion use only, so that we have one place to look at the reserved ids. Where as `PipelineObjectId.fixed("someId")` is for Hyperion users where they want to have a fixed id for some predefined object.

realstraw: But I'm happy to change it to `.fixed("PipelineSchedule")` if it still does not make sense.

src/main/scala/com/krux/hyperion/objects/SnsAlarm.scala
realstraw: but they are already associated with an activity with human readable names I assume

src/main/scala/com/krux/hyperion/objects/Schedule.scala
sethyates: I don't have a strong preference, apart from I had to ask myself what the heck a ScheduleObjectId was and go looking for it, so a more explicit call might be clearer.  But again, I don't have a strong preference either way.

sethyates:
I'm confused by the latest commit.  I was thinking that the changes we had already made met the requirements to fix #37--wasn't suggesting that we should do more.  What do you envisage that the `run` and doing the `runsOn` as a separate parameter set adds?

realstraw:
Just want to show was I was trying to do in #37 if we want a run method. Made a separate commit so we can easily revert.

sethyates:
Oh.  Yeah.  Went over my head.  I thought what we had already done with factoring out all the `run*` and making `runsOn` on the `*Activity` classes was good enough.  Happy to see further what the full idea is behind #37.

sethyates:
I sort of get it.  So you can do something like this:
````scala
val sparkCluster = ...
val activity = sparkCluster.run(SparkActivity()).withSteps(...)
````
as opposed to:
````scala
val sparkCluster = ...
val activity = SparkActivity(sparkCluster).withSteps(...)
````
right?

realstraw:
Yes, actually you can do it in either ways, both
````scala
val sparkCluster = ...
val activity = sparkCluster.run(SparkActivity()).withSteps(...)
````
and
````scala
val sparkCluster = ...
val activity = SparkActivity()(sparkCluster).withSteps(...)
````
are valid, however you need the extra () above as a side effect.

Because the companion function now needs to have two argument groups `def apply(the required args)(runsOn)`, in the SparkActivity example above there is no required arg, hence the extra ()

sethyates:
Hm.  Yeah.  Let's maybe do this in a separate PR then.

realstraw:
Reverted

sethyates:
:+1: :rainbow: LGTM :sparkles: :sparkles: :sparkles: :sparkles: :sparkles: :sparkles:

PR #42, opened 2015-04-15
(fixes #41) refactor opt to seq

### LINKED ISSUE #41: Refactor Seq[T] => Option[Seq[T]]
We have a lot of code in serialize that converts a Seq[T] to Option[Seq[T]] using
```scala
seq match {
  case Seq() => None
  case other => Some(other)
}
```

Refactor them out to a common method/function.

realstraw:


CHANGELOG.md
```diff
@@ -6,7 +6,11 @@ This project adheres to [Semantic Versioning](http://semver.org/).
...
 - TODO
 
-## 1.6.0 - 2015-04-15
+## 1.6.1 - 2015-04-15
+### Changed
+- #41 - Refactor Option to Option[Seq] functions
+
+## 1.6.0 - 2015-04-05
 ### Changed
 - #17 - Make pipline id transaprent
```
sethyates: Typo

src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
```diff
@@ -56,26 +56,11 @@ case class CopyActivity private (
...
-    },
-    onSuccess = onSuccessAlarms match {
-      case Seq() => None
-      case alarms => Some(alarms.map(alarm => AdpRef[AdpSnsAlarm](alarm.id)))
-    },
-    onLateAction = onLateActionAlarms match {
-      case Seq() => None
-      case alarms => Some(alarms.map(alarm => AdpRef[AdpSnsAlarm](alarm.id)))
-    }
+    dependsOn = toOption(dependsOn)(x => AdpRef[AdpActivity](x.id)),
```
sethyates: It would be more awesome if you passed the AdpActivity as a generic type to toOption. It would also be increased awesome if you called it seqToOption.

src/main/scala/com/krux/hyperion/objects/PipelineActivity.scala
```diff
@@ -1,5 +1,7 @@
 package com.krux.hyperion.objects
 
+import com.krux.hyperion.objects.aws.AdpActivity
```
sethyates: Why is this necessary now but not before?

realstraw: Not needed, will remove.

src/main/scala/com/krux/hyperion/objects/CopyActivity.scala
realstraw: > It would be more awesome if you passed the AdpActivity as a generic type to toOption.

Pretty hard to do the "more awesome" part. I had a solution to remove the generic type all together, but it starts this butterfly effect of having to modify too many other things, so decide to give up or save it for later.

> It would also be increased awesome if you called it seqToOption.

will opt for `seqToOption` for the increased awesome.

sethyates: You don't like butterflies???

realstraw: Actually, now you ask. I'll test it a bit further to see what it actually look like.

realstraw:
@sethyates above commits should fix most of the comments

sethyates:
Almost there!  Looking awesome!  Can you give the world a `seqToOption` that does the `(_.ref)` bit?

realstraw:
Tried to do it, but that's a bit hard to do, mainly because we need to make the new `seqToRefOption` return different type based on `def serialize`'s return type.

Unless we want to copy `seqToRefOption` to the sub-classes like `PipelineActivity`, `Database` etc, or introduce parameter type to `PipelineObject` and all it's sub-classes.

sethyates:
:+1: looks awesome to me :sparkles:

realstraw:
:smiley:

PR #47, opened 2015-04-27
Remaining properties on resources

sethyates:


src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
```diff
@@ -8,57 +8,91 @@ import com.krux.hyperion.HyperionContext
...
   sparkVersion: String,
-  terminateAfter: String
+  enableDebugging: Option[Boolean],
+  supportedProducts: Option[String],
+  subnetId: Option[String]
 )(
   implicit val hc: HyperionContext
 ) extends EmrCluster {
 
+  assert(coreInstanceCount >= 1)
```
realstraw: We need at least 2 core nodes for spark clusters

sethyates: Same applies for non-Spark MR jobs?

src/main/scala/com/krux/hyperion/objects/MapReduceCluster.scala
```diff
@@ -8,51 +8,85 @@ import com.krux.hyperion.HyperionContext
...
+  def withTaskInstanceCount(instanceCount: Int) = this.copy(taskInstanceCount = instanceCount)
+  def withTaskInstanceBidPrice(bid: Double) = this.copy(taskInstanceBidPrice = Option(bid))
+  def withKeyPair(keyPair: String) = this.copy(keyPair = Option(keyPair))
+  def withRegion(region: String) = this.copy(region = Option(region))
+  def withDebuggingEnabled() = this.copy(enableDebugging = Option(true))
+  def withSupportedProducts(products: String) = this.copy(supportedProducts = Option(products))
+  def withSubnetId(id: String) = this.copy(subnetId = Option(id))
 
-  val instanceType = hc.emrInstanceType
+  val instanceCount = 1 + coreInstanceCount + taskInstanceCount
```
realstraw: `def` or `lazy val` would be better as `instanceCount` is really required, we are not using it anywhere, but it's there in case it's needed in the application.

src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
```diff
@@ -8,57 +8,91 @@ import com.krux.hyperion.HyperionContext
...
+  def withKeyPair(keyPair: String) = this.copy(keyPair = Option(keyPair))
+  def withRegion(region: String) = this.copy(region = Option(region))
+  def withSparkVersion(sparkVersion: String) = this.copy(sparkVersion = sparkVersion)
+  def withDebuggingEnabled() = this.copy(enableDebugging = Option(true))
+  def withSupportedProducts(products: String) = this.copy(supportedProducts = Option(products))
+  def withSubnetId(id: String) = this.copy(subnetId = Option(id))
 
-  val bootstrapAction = s"s3://support.elasticmapreduce/spark/install-spark,-v,$sparkVersion,-x" ::
-      hc.emrEnvironmentUri.map(env => s"${hc.scriptUri}deploy-hyperion-emr-env.sh,$env").toList
+  val instanceCount = 1 + coreInstanceCount + taskInstanceCount
```
realstraw: I think `def` or `lazy val` is better in this case

src/main/scala/com/krux/hyperion/objects/SparkCluster.scala
realstraw: That I'm not sure, I vaguely remember if you don't have task node, then 1 core node seems to be fine, but if you have task nodes, then you need 2 core nodes.

realstraw:

sethyates:
Hey man, thanks for the review and :sparkles:.  Will merge once Travis goes :green_heart: .

PR #55, opened 2015-05-06
Calling with* when a sequence appends to existing sequence (in most cases)

### LINKED ISSUE #54: The with* methods on sequences should append to sequences
For example, if I have a `SparkActivity` and call `withSteps` multiple times, then the earlier ones are discarded.  I would expect the calls to be additive.

sethyates:
Fixes #54

### DIFF
--- a/src/main/scala/com/krux/hyperion/objects/CsvDataFormat.scala
+++ b/src/main/scala/com/krux/hyperion/objects/CsvDataFormat.scala
@@ -7,18 +7,17 @@ import com.krux.hyperion.objects.aws.AdpCsvDataFormat
  */
 case class CsvDataFormat private (
   id: PipelineObjectId,
-  column: Seq[String],
+  columns: Seq[String],
   escapeChar: Option[String]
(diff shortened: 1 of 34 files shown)


src/main/scala/com/krux/hyperion/objects/CsvDataFormat.scala
```diff
@@ -11,8 +11,8 @@ case class CsvDataFormat private (
   escapeChar: Option[String]
 ) extends DataFormat {
 
+  def withColumn(col: String*) = this.copy(column = column ++ col)
   def withColumns(cols: Seq[String]) = this.copy(column = cols)
```
realstraw: Do we still need the `Seq` version at all? I recon we can remove these ones

sethyates: Not sure.  I'd happily do that.  You can convert a Seq into the var args using `: _*`.  I'll make that change here.

realstraw:

PR #57, opened 2015-05-20
(fixes #56) reorganize objects and group them into packages

### LINKED ISSUE #56: Re-organise objects and group them by activities and resources for easy navigation
Currently everything is under `com.krux.hyperion.objects`. As number of activities grow, it's better to organise them into `activities`, `resources`, etc. for easier navigation.

realstraw:


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ## 1.8.0 - unreleased
 ### Changed
 - #54 - with* methods that take a sequence are now additive, and replaced withColumns(Seq[String]) with withColumns(String...)
+- #56 - reorganize objects into packages by type
```
sethyates: Hm. Didn't we actually already release 1.8.0?

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
```diff
@@ -4,6 +4,11 @@ import com.krux.hyperion.objects._
 import com.krux.hyperion.Implicits._
 import com.krux.hyperion.DataPipelineDef
 import com.krux.hyperion.HyperionContext
+import com.krux.hyperion.objects.activity.RedshiftCopyActivity
+import com.krux.hyperion.objects.database.RedshiftDatabase
```
sethyates: So this introduces an inconsistency. The namespace has plural objects then the next entry is singular (activity, database). I think we should drop the "objects" namespace and pluralise the others like "activities" etc.

src/main/scala/com/krux/hyperion/objects/activity/HadoopActivity.scala
```diff
@@ -1,14 +1,18 @@
...
 import com.krux.hyperion.objects.aws.AdpEmrActivity
+import com.krux.hyperion.objects.precondition.Precondition
+import com.krux.hyperion.objects.resource.EmrCluster
 
 /**
- * Defines a MapReduce activity
+ * Runs hadoop steps on an Amazon EMR cluster
  */
-case class MapReduceActivity private (
+case class HadoopActivity private (
```
sethyates: Why the name change?

src/main/scala/com/krux/hyperion/objects/activity/SqlActivity.scala
```diff
@@ -1,11 +1,20 @@
...
 
+/**
+ * Runs an SQL query on a RedShift cluster. If the query writes out to a table that does not exist,
+ * a new table with that name is created.
+ */
 case class SqlActivity private (
   id: PipelineObjectId,
   runsOn: Ec2Resource,
-  database: Database,
+  database: RedshiftDatabase,
```
sethyates: Is this specific to Redshift? I thought it worked for any Database.

src/main/scala/com/krux/hyperion/objects/dataformat/TsvDataFormat.scala
```diff
@@ -1,9 +1,11 @@
-package com.krux.hyperion.objects
+package com.krux.hyperion.objects.dataformat
 
+import com.krux.hyperion.objects.PipelineObjectId
 import com.krux.hyperion.objects.aws.AdpTsvDataFormat
 
 /**
- * TSV data format
+ * A comma-delimited data format where the column separator is a tab character and the record
```
sethyates: Well technically it isn't comma-delimited. It is tab-delimited. Or delimited with a tab.

src/main/scala/com/krux/hyperion/objects/parameter/Parameter.scala
```diff
@@ -0,0 +1,14 @@
+package com.krux.hyperion.objects.parameter
+
+import com.krux.hyperion.objects.aws.AdpParameter
+
+trait Parameter {
+  def id: String
+  def description: Option[String]
+  def encrypted: Boolean
+
+  val name = if (encrypted) s"*my$id" else s"my$id"
```
sethyates: Should probably be lazy val or def, right?

CHANGELOG.md
realstraw: Not in here: https://oss.sonatype.org/content/repositories/releases/com/krux/hyperion_2.10/
I thought you did something to make it not releasing 1.8.0

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
realstraw: OK, sounds good

CHANGELOG.md
sethyates: Oh okay.

src/main/scala/com/krux/hyperion/objects/activity/HadoopActivity.scala
realstraw: Because I wanted to make it look more different than EmrActivity. And Spark / Hadoop seems more at the same level.

src/main/scala/com/krux/hyperion/objects/activity/SqlActivity.scala
realstraw: That's what I thought as well, but from the aws doc, it seems it's redshift only...

http://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-object-sqlactivity.html

src/main/scala/com/krux/hyperion/objects/dataformat/TsvDataFormat.scala
realstraw: Agreed, it's copy pasted from aws doc. Will change.

src/main/scala/com/krux/hyperion/objects/activity/HadoopActivity.scala
sethyates: Ok. I'm not hung up either way but I consciously chose MapReduceActivity before because technically Hive and Pig run on Hadoop (our Spark also runs on Hadoop YARN, right?). Anyway more a question than a blocker.

src/main/scala/com/krux/hyperion/objects/activity/SqlActivity.scala
sethyates: Dang!

src/main/scala/com/krux/hyperion/objects/parameter/Parameter.scala
realstraw: Yes

src/main/scala/com/krux/hyperion/examples/ExampleRedshiftLoad.scala
realstraw: Dropped `objects`. Shouldn't we keep the others in singlular? It seems to be the package naming convention in Java and Scala, like `scala.collection` and `java.util`.

sethyates: Resources are at different places in the folder hierarchy.

There is no actual standard governing past the domain name. Quoting from the Oracle doc:

http://www.oracle.com/technetwork/java/codeconventions-135099.html

It is whatever reads the clearest. at least the inconsistency has been fixed by dropping objects. I'll leave to you on whether to pluralise or not, but have a look at it both ways.

realstraw:
@sethyates above commit changed the package names, revert Hadoop* back to MapReduce*, and fixed some of the wording in comments.

src/main/scala/com/krux/hyperion/activity/DeleteS3PathActivity.scala
```diff
@@ -56,7 +60,7 @@ case class DeleteS3PathActivity private (
 
 }
 
-object DeleteS3PathActivity {
+object DeleteS3PathActivity extends RunnableObject {
```
sethyates: Why is the companion object extending RunnableObject and not the case class?

src/main/scala/com/krux/hyperion/activity/GoogleStorageActivity.scala
```diff
@@ -129,7 +134,7 @@ case class GoogleStorageUploadActivity private (
 
 }
 
-object GoogleStorageUploadActivity {
+object GoogleStorageUploadActivity extends RunnableObject {
```
sethyates: Why is the companion object extending RunnableObject?

sethyates:
I don't get why the companion objects instead of the classes themselves are extending RunnableObject

src/main/scala/com/krux/hyperion/activity/GoogleStorageActivity.scala
realstraw: We already had these for existing SparkActivity and MapReduceActivity object, this change is mainly for consistency. It cannot be extended by the class as we need them before the class is initialised.

e.g.
https://github.com/krux/hyperion/blob/master/src/main/scala/com/krux/hyperion/examples/ExampleSpark.scala#L51
https://github.com/krux/krux-spark-analytics/blob/master/krux-emr-jobs/src/main/scala/com/krux/emr/pipelines/MonthlyClientSegmentPublisherOverlap.scala#L53

Having said that, this is not ideal neither, since now `GoogleStorageUpLoadActivity.scheduledStartTime` is no different from `SparkActivity.scheduledStartTime`. Better one may simply be making `RunnableObject` itself a object instead of a trait, but we could discuss/fix it in a separate issue.

sethyates:
LGTM :sparkles:

PR #61, opened 2015-06-17
(fixes #59) add new workflow DSL

### LINKED ISSUE #59: Implement a dependency graph DSL to be returned by workflow in PipelineDef
Instead of returning the activity that is the last one in the dependency graph, design and implement a DSL that connects all activities together at the end for dependencies.
e.g. instead of having
```scala
def workflow = {
  val a1 = Activity1()
  val a2 = Activity2().dependsOn(a1)
  val a3 = Activity3().dependsOn(a1)
  val a4 = Activity4().dependsOn(a2, a3)

  Some(a4)
}
```
change it to something like:
```scala
def workflow = {
  val a1 = Activity1()
  val a2 = Activity2()
  val a3 = Activity3()
  val a4 = Activity4()
  
  a1 ~> a2,
        a3 ~> a4
}
```

realstraw:


CHANGELOG.md
```diff
@@ -2,11 +2,14 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## 1.8.0 - unreleased
+## 1.8.0 - 2015-06-17
```
sethyates: I think we've (or I've) lost track of the versions in the CHANGELOG somewhere as I see the last master build in Travis was 1.8.1.

src/main/scala/com/krux/hyperion/WorkflowDSL.scala
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  def ~>(act: PipelineActivity): Seq[PipelineActivity] = {
+    assertNoDependencies(act)
+    Seq(act.dependsOn(activities:_*))
+  }
+
+  def andThen(act: PipelineActivity): Seq[PipelineActivity] = this.~>(act)
+
+  def ~>(acts: Seq[PipelineActivity]): Seq[PipelineActivity] = {
```
sethyates: Can you also do a `<~` which basically operates in the opposite direction, so `A <~ B` would be the same as `B ~> A`?

src/main/scala/com/krux/hyperion/WorkflowDSL.scala
```diff
@@ -0,0 +1,50 @@
...
+    assertNoDependencies(acts:_*)
+    for (act <- acts) yield act.dependsOn(activities:_*)
+  }
+
+  def andThen(acts: Seq[PipelineActivity]): Seq[PipelineActivity] = this.~>(acts)
+
+  def +(act: PipelineActivity): Seq[PipelineActivity] = activities :+ act
+
+  def and(act: PipelineActivity): Seq[PipelineActivity] = this.+(act)
```
sethyates: Nit: Personally, I think it would be more readable if the implementations were swapped like so:
````scala
def +(act: PipelineActivity): Seq[PipelineActivity] = this.and(act)

def and(act: PipelineActivity): Seq[PipelineActivity] = activities :+ act
````
in all cases where there is an operator.

Comment (realstraw) [inline src/main/scala/com/krux/hyperion/WorkflowDSL.scala]: Make sense. I'll change

File: src/main/scala/com/krux/hyperion/WorkflowDSL.scala (hunk 2 above)
Comment (realstraw) [inline src/main/scala/com/krux/hyperion/WorkflowDSL.scala]: Seems it's just dependsOn, maybe we can add `<~` in all activities that is the alias of `dependsOn`?

Comment (sethyates) [inline src/main/scala/com/krux/hyperion/WorkflowDSL.scala]: Not sure this is a real use case, but what if you had `(A + B) <~ C`?

File: CHANGELOG.md (hunk 1 above)
Comment (sethyates) [inline CHANGELOG.md]: Also need to update the version in `build.sbt`.

File: src/main/scala/com/krux/hyperion/WorkflowDSL.scala (hunk 2 above)
Comment (realstraw) [inline src/main/scala/com/krux/hyperion/WorkflowDSL.scala]: Ah, yes then in that case the `dependsOn` alias would not work, it needs to be in here.

But do we want to support that though? Seems a bit weird to me to have `A <~ B`, and if people start mixing `~>` and `<~` (e.g. `A ~> B <~ C ~> D`), it might be a bit confusing.

Comment (sethyates) [inline src/main/scala/com/krux/hyperion/WorkflowDSL.scala]: "With great power comes great responsibility" - Uncle Ben ;-)

Comment (realstraw) [inline src/main/scala/com/krux/hyperion/WorkflowDSL.scala]: Hmm, it's harder than I thought it would be. Implemented it like this
```scala
  def priorTo(act: PipelineActivity): Seq[PipelineActivity] =
    activities.map(_.dependsOn(act))
  def <~(act: PipelineActivity): Seq[PipelineActivity] = this.priorTo(act)
```
but this make the following two inconsistent
```scala
// This one is correct
act6 <~ ((act4 + act5) <~ ((act2 + act3) <~ act1))
// This one is not correct
act6 <~ (act4 + act5) <~ (act2 + act3) <~ act1
```

Tried some other small variations, seems it's non-trivial to support `<~` without unambiguous.

Maybe we have to give up the great power and take less responsibilities ;)

sethyates: Ah, I think it needs to be right-associative, so it'll probably have to be `<~:`

realstraw: And `priorTo` becomes `priorTo_:`. Looks a bit ugly

realstraw:
@sethyates all comments fixed

sethyates:
LGTM :sparkling_heart:

PR #67, opened 2015-07-02
(Fixes #64) Supports non-default region

### LINKED ISSUE #64: Ability to create pipelines in non-default region

sethyates:


realstraw:
Should we make the client use `hyperion.aws.access_key_id` and `hyperion.aws.access_key_secret` as well?

sethyates:
I think the default credentials provider chain is fine (apart from maybe using STS to assume role). If anything, this should be specified from outside the app/configuration for security reasons.

sethyates:
Plus it would be a bit difficult to do this reliably given the workflow can override configuration.

realstraw:
OK :+1:

Issue #71, opened 2015-07-02
Support passing Parameters


sethyates:
Currently, PipelineObjects take parameter types in a typesafe manner.  This precludes using Parameters for the non-string values.  In order to support a wider usage of Parameters without sacrificing type safety, we should implement a `ParameterOf[T]` class that is implicitly convertable from `T` and `Parameter` and emits the correct string value (e.g., Parameters will be `#{name}`).  Update all relevant parameters to take `ParameterOf[T]` instead of `T`.

* `ParameterOf[String]` will be convertable from `String` and `StringParameter`.
* `ParameterOf[Int]` will be convertable from `Int` and `IntegerParameter`.
* `ParameterOf[Double]` will be convertable from `Double` and `DoubleParameter`.
* and so on

sethyates:
Fixed in 2.0

PR #72, opened 2015-07-02
POC for ParameterOf[T] for issue #71

sethyates:
Proof-of-concept for feedback.


realstraw:
The approach looks good. My concern is that it seems it's a lot of work to support parameters without much gains. Also parameter itself have many constraints like prefix "my" and it needs a specific format. Do we have any use case that parameters have to be used?

sethyates:
Well my thought is that some things we might want to change dynamically without changing the structure of the pipeline, and that might be easier via parameters.  Thinking about maybe spot prices for example.

realstraw:
Parameters cannot be changed dynamically without recompile when using our cli to submit and activate jobs neither right? Are you thinking using the cli to generate parameterised json templates and manage them through other means?

sethyates:
Wasn't thinking of it via the CLI, but through other means.

PR #75, opened 2015-07-06
(Fixes #74) DataNode should return path using toString

### LINKED ISSUE #74: DataNode should return path using toString

sethyates:


ysalomon:
:+1:

PR #79, opened 2015-07-14
(Fixes #78) Strip trailing $ from MainClass

### LINKED ISSUE #78: Strip trailing $ from MainClass
Calling `.withMainClass(SomeClass.getClass.getName)` results in some cases with a Scala object with the trailing $ which doesn't work so well when running as a jar.  Strip the trailing $ in cases where we set the MainClass:
* JarActivity
* MapReduceStep
* SparkStep

sethyates:


src/main/scala/com/krux/hyperion/activity/MapReduceStep.scala
```diff
@@ -4,15 +4,15 @@ package com.krux.hyperion.activity
...
-  jar: String = "",
-  mainClass: String = "",
+  jar: Option[String] = None,
+  mainClass: Option[String] = None,
   args: Seq[String] = Seq()
 ) {
 
-  def withJar(jar: String) = this.copy(jar = jar)
-  def withMainClass(mainClass: String) = this.copy(mainClass = mainClass)
+  def withJar(jar: String) = this.copy(jar = Option(jar))
```
realstraw: Why are we making them optional?

sethyates: Semantically more correct than empty string.  Plus it has the nice `.toSeq` below.

sethyates: And more consistent to the way we're doing it everywhere else.

realstraw: I think we should make them with no default empty value and make them required. We require it to be specified in the companion object any way

sethyates: > We require it to be specified in the companion object any way
No we don't.  These are setup to be specified via `withJar` and `withMainClass`.  `mainClass` strictly isn't required if the jar has the correct manifest setup.  So the only one required is `jar`.

> I think we should make them with no default empty value and make them required. 
OK well we can discuss that and I have no objection to it, but that's a breaking change and orthogonal to this PR.

sethyates: Add to this, we've got several other places already where strictly required values are marked as Option and the caller is required to fill in using `with*` (`withSteps` being the first to come to mind).

sethyates: All that being said, I'm happy to revert that bit if it is contentious, since it is secondary to my main change here.

realstraw: Ah yes. Didn't see very clear on the phone.

realstraw:
Can we also have withClass that takes a class object and get its string form?

sethyates:
Hehe, I was already working on that :-)

realstraw:
:+1:

PR #82, opened 2015-07-15
(Fixes #81) Fix call to getCurrentRegion to default to US-EAST-1

### LINKED ISSUE #81: Regression introduced by #64 requires region to be specified if not running in EC2
A side effect of #64 is that the --region parameter is now required on non-EC2 instances due to the call to `getCurrentRegion`.

sethyates:


realstraw:
:+1:

PR #86, opened 2015-07-15
(Fixes #80) Change jar-based activities/steps to require a jar

### LINKED ISSUE #80: Change jar-based activities/steps to require a jar
The following require a jar:
* JarActivity
* MapReduceStep
* SparkStep

This is currently specified via `withJar` but should really be required in the constructor/companion.

sethyates:


realstraw:
:+1:

PR #87, opened 2015-07-15
(Fixes #84) Drop dependsOn and require WorkflowDSL

### LINKED ISSUE #84: Drop dependsOn and require WorkflowDSL

sethyates:


realstraw:
Can you update example as well. IIRC we still have one or two examples that use depndsOn

sethyates:
All of the examples already use the DSL.

realstraw:
Ah ok :+1:

PR #93, opened 2015-07-19
Refactored into a 'core' project

sethyates:


build.sbt
```diff
@@ -75,14 +87,21 @@ lazy val commonSettings = Seq(
 
 lazy val root = (project in file(".")).
   settings(commonSettings: _*).
+  settings(name := "hyperion").
+  dependsOn(core)
```
realstraw: What is dependsOn? This should be aggregate.

sethyates: No it shouldn't (or it needs to be in addition).  Without the `dependsOn`, the dependency doesn't appear in the aggregate pom.

sethyates: Still behaves as if `aggregate` were there:
````
$ sbt compile
[info] Loading project definition from /Users/seth/src/hyperion/project
[info] Set current project to hyperion (in build file:/Users/seth/src/hyperion/)
[info] Updating {file:/Users/seth/src/hyperion/}core...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] Updating {file:/Users/seth/src/hyperion/}root...
[info] Resolving jline#jline;2.12.1 ...
[info] Done updating.
[info] Compiling 94 Scala sources to /Users/seth/src/hyperion/core/target/scala-2.11/classes...
[success] Total time: 30 s, completed 19/07/2015 10:18:09 AM
````

sethyates:
I'm going to move this to a 2.0 branch.

build.sbt
realstraw: You sure? I thought there's something that you need for root to use aggregate, and also use dependsOn if you want a aggregated jar. But I cannot remember why aggregate is needed though. It might be changed so that only dependsOn is needed

sethyates: Yeah I'm sure.  http://www.scala-sbt.org/0.12.2/docs/Getting-Started/Multi-Project.html

`aggregate` just means that multiple projects are aggregated together for commands like `compile`, so if you do `compile` at the root level, it will `compile` all aggregated projects.  It doesn't express a dependency and won't put that dependency in the published artifact

`dependsOn` on the other hand expresses a dependency which will make its way into the published artifact.  It will also ensure the depended-on project is compiled before the project.

PR #100, opened 2015-07-23
Hyperion CLI driver should exit with appropriate error codes

sethyates:


scripts/run-jar.sh
```diff
@@ -19,4 +19,4 @@ fi
 
 # Run the jar itself.
 echo "Running jar ${LOCAL_JAR} $@"
-exec java -cp ${LOCAL_JAR} $@
+echo java -cp ${LOCAL_JAR} $@ | exec sh
```
realstraw: Why is this better?

sethyates: Because it actually works

realstraw:
:+1:

Issue #101, opened 2015-07-23
Allow workflow DSL to have duplicated activities for flexibility


realstraw:
e.g.
Allow `(act1 :~> act2) + (act3 :~> act2)` and this should be equivalent as `(act1 + act3) :~> act2`

PR #102, opened 2015-07-23
(fixes #91 #101) workflow dsl now function as expected

### LINKED ISSUE #91: workflow dsl broken when the right hand side of andThen have dependencies
e.g.
```scala
act1 :~> ((act2 :~> act3) + (act4 :~> act5))
```

returns act2 does not depend on anything

realstraw:


src/main/scala/com/krux/hyperion/WorkflowExpression.scala
```diff
@@ -0,0 +1,38 @@
+package com.krux.hyperion
+
+import com.krux.hyperion.activity.PipelineActivity
+import com.krux.hyperion.common.PipelineObject
+
+sealed abstract class WorkflowExpression {
```
sethyates: Why is this better than `trait`?

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
```diff
@@ -0,0 +1,38 @@
...
+
+  def toPipelineObjects: Iterable[PipelineActivity] = {
+
+    def toPipelineObjectsRec(exp: WorkflowExpression): Set[PipelineActivity] = {
```
sethyates: `{` here not necessary.

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
```diff
@@ -0,0 +1,38 @@
...
+      exp match {
+        case WorkflowActivityExpression(act) => Set(act)
+        case WorkflowArrowExpression(left, right) =>
```
sethyates: Space before helps readability.

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
```diff
@@ -0,0 +1,38 @@
...
+          val leftDeps = toPipelineObjectsRec(left)
+          val rightDeps = toPipelineObjectsRec(right)
+          // rightDeps now should depend on leftDeps
+          rightDeps.map(_.dependsOn(leftDeps.toSeq:_*)) ++ leftDeps
+        case WorkflowPlusExpression(left, right) =>
```
sethyates: Space before helps readability.

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
```diff
@@ -0,0 +1,38 @@
...
+          val leftDeps = toPipelineObjectsRec(left)
+          val rightDeps = toPipelineObjectsRec(right)
+          (leftDeps ++ rightDeps).groupBy(_.id)
+            .map { case (id, acts) =>
+              acts.reduceLeft((a, b) => a.dependsOn(b.dependsOn:_*))
```
sethyates: Why are you doing this?  The plus operation expresses no dependency.  Shouldn't this just be building a set of left and right?

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
realstraw: it'll make the compiler check the completeness when doing match. It will give an warning if you are missing a case, like when you match `Option` or `List`.

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
realstraw: This is for the case that you might want duplicated activity in the expression (test case https://github.com/krux/hyperion/pull/102/files#diff-d837e1f66fd536197f3599b615e546fcR63). E.g. `(a :~> b) + (c :~> b)` will return the correct dependency, but not when we simply union left and right.

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
sethyates: TIL. Excellent, thanks.

src/main/scala/com/krux/hyperion/WorkflowExpression.scala
sethyates: I see.  Thanks!

sethyates:
:sparkles::sparkles: :+1: LGTM :rainbow: :sparkles: :sparkles:

PR #104, opened 2015-07-27
[2.0] Work on scripts and file activities

sethyates:


contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileMerger.scala
```diff
@@ -0,0 +1,49 @@
...
+    if (source.getName == "-") {
+      print("Merging stdin...")
+      IOUtils.copy(doSkipFirstLine(System.in), output)
+      println("done")
+    } else {
+      print(s"Merging ${source.getAbsolutePath}...")
+      val input = Option(new FileInputStream(source))
+        .map(s => if (source.getName.endsWith(".gz")) new GZIPInputStream(s) else s)
+        .map(s => new BufferedInputStream(s))
+        .get
```
realstraw: Why wrap with an `Option` and then perform a `get` at the end? I might look simpler to do:
```scala
val input = new BufferedInputStream({
  val s = new FileInputStream(source)
  if (s.getName.endsWith(".gz")) GZIPInputStream(s)
  else s
})
```

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileRepartitioner.scala
```diff
@@ -0,0 +1,85 @@
...
+
+case class FileRepartitioner(options: Options) {
+
+  def repartition(): Boolean = moveFiles(nameFiles(split(merge())))
+
+  private def merge(): File = options.inputs match {
+    case Seq(one) => one
+
+    case files =>
+      val destination: File = File.createTempFile("merge-", ".tmp", options.temporaryDirectory.get)
```
realstraw: This will throw an exception when `options.temporaryDirectory` is `None`. Is it intended?

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileMerger.scala
sethyates: Thought it looked cleaner than that option (pun intended).  Could have used `Seq`/`head` as well.

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileRepartitioner.scala
sethyates: By this point, `options.temporaryDirectory` is not `None` and yes.

sethyates: The option could be the repetitive:
```
options.temporaryDirectory.map(dir => File.createTempFile("merge-", ".tmp", dir)).getOrElse(File.createTempFile("merge-", ".tmp"))
```

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileRepartitioner.scala
```diff
@@ -0,0 +1,85 @@
...
+            Files.move(source, dest, StandardCopyOption.REPLACE_EXISTING)
+        }
+      } else if (options.link) {
+        Files.createSymbolicLink(dest, source)
+      } else {
+        Files.copy(source, dest, StandardCopyOption.REPLACE_EXISTING)
+      }
+    }
+
+    true
```
realstraw: It seems this `forall` will always return `true`.

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileSplitter.scala
```diff
@@ -0,0 +1,86 @@
+package com.krux.hyperion.contrib.activity.file
+
+import java.io._
+import java.util.zip.GZIPOutputStream
+import org.apache.commons.io.IOUtils
+
+import scala.collection.mutable.ListBuffer
+
+case class FileSplitter(
```
realstraw: I think it's better to have stateful classes as normal class instead of case class

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/RepartitionFile.scala
```diff
@@ -0,0 +1,158 @@
...
+        val matcher = options.pattern.map(pattern => FileSystems.getDefault.getPathMatcher(s"glob:$pattern"))
+
+        override def accept(dir: File, name: String): Boolean = matcher.forall(_.matches(Paths.get(name)))
+      }).toSeq
+    } else if (k.isFile) {
+      if (k.canRead) {
+        Seq(k.getAbsoluteFile)
+      } else {
+        System.err.println(s"ERROR: Cannot read $k")
+        Seq()
```
realstraw: Not sure in your case, but in many cases the compiler has trouble infer the return type of the whole expression when you have `Seq()` like this, and results in the type being `Seq[Nothing]`. We normally do `Seq.empty[File]` or `Seq[File]()` to explicitly specify the generic type to avoid this error.

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/RepartitionFile.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  def applyDefaults(options: Options): Options =
+    Seq(
+      applyDefaultTemporaryDirectory _,
+      applyDefaultInputs _,
+      applyDefaultDirectory _,
+      applyDefaultNumberOfFiles _,
+      applyDefaultFileChecks _,
+      applyDefaultNumberOfFilesCalculation _
+    ).foldLeft(options) { case (acc, handler) => handler(acc) }
```
realstraw: no need `case` here `.foldLeft(options)((acc, handler) => handler(acc))` should just work.

sethyates: Ah yes that's right.  Will change.

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/RepartitionFile.scala
sethyates: Seemed to have no problem.

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileSplitter.scala
sethyates: True.  Will change.

contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileRepartitioner.scala
sethyates: True.

PR #105, opened 2015-07-31
2.0

sethyates:


build.sbt
```diff
@@ -58,11 +67,18 @@ site.includeScaladoc()
...
 lazy val commonSettings = Seq(
   organization := "com.krux",
-  scalacOptions ++= Seq("-deprecation", "-feature", "-Xlint", "-Xfatal-warnings"),
   version := hyperionVersion,
   scalaVersion := "2.11.6",
-  crossScalaVersions := Seq("2.10.4", "2.11.6"),
-  libraryDependencies += scalatestArtifact,
+  crossScalaVersions := Seq(
+    "2.10.4",
+    "2.11.6"
```
realstraw: should we change this to `2.11.7`?

core/src/main/scala/com/krux/hyperion/activity/HadoopActivity.scala
```diff
@@ -0,0 +1,100 @@
+package com.krux.hyperion.activity
+
+import com.krux.hyperion.action.SnsAlarm
+import com.krux.hyperion.aws._
+import com.krux.hyperion.common.{PipelineObject, PipelineObjectId}
+import com.krux.hyperion.precondition.Precondition
+import com.krux.hyperion.resource.{WorkerGroup, EmrCluster}
+
+case class HadoopActivity(
```
realstraw: I think it's better to provide a short description here to show the difference between this and MapRedcueActivity, so that people looking at code don't have to check the docs in amazon website.

build.sbt
sethyates: Was thinking that as well.  Will make the change.

core/src/main/scala/com/krux/hyperion/activity/HadoopActivity.scala
sethyates: Done

core/src/main/scala/com/krux/hyperion/WorkflowExpression.scala
```diff
@@ -0,0 +1,63 @@
...
+    toPipelineObjectsRec(this)
+  }
+
+  def andThen(right: WorkflowExpression): WorkflowExpression = WorkflowArrowExpression(this, right)
+  def :~>(right: WorkflowExpression): WorkflowExpression = this.andThen(right)
+  def ~>(right: WorkflowExpression): WorkflowExpression = this.andThen(right)
+
+  def priorTo_:(right: WorkflowExpression): WorkflowExpression = this.andThen(right)
+  def <~:(right: WorkflowExpression): WorkflowExpression = this.priorTo_:(right)
+  def <~(right: WorkflowExpression): WorkflowExpression = this.priorTo_:(right)
```
realstraw: remove `:` would make it non-right associative, which may produce incorrect workflow, (at least last time it did)

sethyates: Looks like with recent workflow changes, it doesn't make a difference.  I added a unit test that produces the same dependencies with both.

realstraw: so `<~` should be `priorTo` or `priorTo_:`?

realstraw:
Let's freeze the refactor a bit and leave them for later releases, it become very hard to view and track any new changes given the size of the PR.

I think after we removed `priorTo`, `<~:` and `:~>` in favor of `~>` only, we should be good to go now.

sethyates:
Yeah agreed. <~: and friends removal coming shortly.

sethyates:
OK changes have been made.  Consider this frozen.  I've tested several pipelines (generation) and compared to previous and it checks out, but I'd recommend checking any pipelines you might be interested in.

realstraw:
:+1: :sparkles: :sparkles: :sparkles: :fireworks: :rainbow: Let's do this!

PR #124, opened 2015-08-02
Support sets on WorkflowExpression

sethyates:


src/main/scala/com/krux/hyperion/WorkflowDSL.scala
```diff
@@ -21,11 +21,20 @@ class WorkflowDSL(left: WorkflowExpression) {
 
 object WorkflowDSL {
 
-  implicit def activity2WorkflowDSL(act: PipelineActivity): WorkflowDSL =
-    new WorkflowDSL(WorkflowActivityExpression(act))
+  implicit def activitySet2WorkflowExpression(act: Set[PipelineActivity]): WorkflowExpression =
```
realstraw: not that it matters much, but I think it's better to change `act` to `acts` when it's a collection

sethyates:
Subsumed with feedback into #105

PR #134, opened 2015-08-04
(Fixes #133) SftpActivity needs to support S3 URLs for identity file and download as appropriate

### LINKED ISSUE #133: SftpActivity needs to support S3 URL's for identity file and download as appropriate

sethyates:


contrib/activity/sftp/src/main/scala/com/krux/hyperion/contrib/activity/sftp/SftpActivity.scala
```diff
@@ -55,13 +59,40 @@ object SftpActivity {
...
+          case bucket :: key =>
+            print(s"Downloading identity from s3://$bucket/${key.mkString("/")}...")
+            val s3Object = Try(new AmazonS3Client().getObject(bucket, key.mkString("/")))
+            if (s3Object.isFailure) {
+              None
+            } else {
+              println("done.")
+              print("Attaching identity...")
+              s3Object.toOption
+            }
```
realstraw: It looks better if we use case match `Success` and `Failure` here

sethyates: Ha! Great point. Will change.

realstraw:
:+1:

PR #139, opened 2015-08-06
(fixes #138) make parameter key to be able to use lower case starting letter

### LINKED ISSUE #138: Make parameter name always work
AWS datapipeline shows the following error when the letter after my is in lower case

    ERROR: Parameter id must have prefix 'my'

e.g. it works with `myAwsParam`, but it gives above error if it is `myawsParam` it will give the above un-helpful message.

We should either alert before submit pipeline, or automatically change the lower case to upper case automatically

realstraw:


sethyates:
:+1: :sparkles: :rainbow:

PR #151, opened 2015-08-14
(Fixes #149, Fixes #150) DataNodes return wrong types and don't include preconditions and actions in objects

### LINKED ISSUE #149: Preconditions are not returned in objects for DataNodes

### LINKED ISSUE #150: The `whenMet` method returns DataNode instead of S3DataNode

sethyates:


core/src/main/scala/com/krux/hyperion/datanode/RedshiftDataNode.scala
```diff
@@ -22,18 +22,18 @@ case class RedshiftDataNode private (
   onFailAlarms: Seq[SnsAlarm]
 ) extends DataNode {
 
-  def named(name: String) = this.copy(id = PipelineObjectId.withName(name, id))
-  def groupedBy(group: String) = this.copy(id = PipelineObjectId.withGroup(group, id))
+  def named(name: String): RedshiftDataNode = this.copy(id = PipelineObjectId.withName(name, id))
```
realstraw: why are we adding all these return types explicitly but removing the explicit return types above?

sethyates: Because the trait has these methods defined returning the trait type and so if you use these methods without specifying the sub class's type you can't properly chain the calls. With regard to removing the type on toString, this isn't changing the return type and so the explicit type isn't required.

realstraw:

PR #152, opened 2015-08-14
(Fixes #150) The `whenMet` method returns DataNode instead of S3DataNode

### LINKED ISSUE #150: The `whenMet` method returns DataNode instead of S3DataNode

sethyates:


core/src/main/scala/com/krux/hyperion/datanode/S3DataNode.scala
```diff
@@ -16,11 +16,16 @@ sealed trait S3DataNode extends Copyable {
   def asOutput(): String = asOutput(1)
   def asOutput(n: Integer): String = "${" + s"OUTPUT${n}_STAGING_DIR}"
 
-  def withDataFormat(fmt: DataFormat): S3DataNode
-
   def named(name: String): S3DataNode
   def groupedBy(client: String): S3DataNode
 
+  def withDataFormat(fmt: DataFormat): S3DataNode
```
realstraw: I'm not sure, but does it work if you change the `whenMet` in `DataNode` to
````scala
def whenMet(conditions: Precondition*): this.type
```

sethyates: I'll give it a go.

sethyates: Doesn't seem very joyous.  Can I defer that for later?

realstraw:

PR #154, opened 2015-08-17
(Fixes #153) The create --force action doesn't detect existing pipelines…

### LINKED ISSUE #153: The create --force action doesn't detect existing pipelines if there are more than 25 active pipelines

sethyates:
…if there are more than 25 active pipelines

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
+++ b/core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
@@ -1,11 +1,12 @@
 package com.krux.hyperion
 
+import scala.annotation.tailrec
+import scala.collection.JavaConversions._
 import com.amazonaws.auth.{DefaultAWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
 import com.amazonaws.regions.{Regions, Region}
 import com.amazonaws.services.datapipeline._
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
```diff
@@ -35,12 +36,23 @@ class HyperionAwsClient(regionId: Option[String] = None, roleArn: Option[String]
...
 
     def getPipelineId: Option[String] = {
-      val idNameToIdPartial = Function.unlift((idName: PipelineIdName) =>
-        if (idName.getName == pipelineName) Option(idName.getId) else None
-      )
+      @tailrec
+      def queryPipelines(ids: List[String] = List(), request: ListPipelinesRequest = new ListPipelinesRequest()): List[String] = {
+        val response = client.listPipelines(request)
+
+        val theseIds: List[String] = response.getPipelineIdList.collect({
```
realstraw: I don't think you need `(` here, only `{` should be enough.

realstraw:
:+1:

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
sethyates: True.

PR #156, opened 2015-08-17
(fixes #155) workflow returns the wrong dependency

### LINKED ISSUE #155: Workflow breaks when having ArrowDependency on the right hand side
e.g. `act1 ~> (act2 ~> act3)` returns pipeline not having `act2` depend on `act1`

realstraw:
- when right hand side have arrow depenncies


ysalomon:
:+1:

PR #165, opened 2015-08-27
(fixes #157) uses a separate workflow/dependency graph to manage dependency building

### LINKED ISSUE #157: Re-examine how we do dependencies
The fix of https://github.com/krux/hyperion/pull/155 introduces in consistent behavior to pipeline activities where most other methods returns a copy of the object but dependsOn does not. Examine and act on making it consistent again. Potential solutions:

- Decouple dependsOn from activity objects and make it part of workflow expression.
- Workflow performs a "lazy build", and only performs one dependsOn per object.

Also the way we fetch all involved objects in workflow now becomes outdated, as we are not returning the last object that have all the dependencies in workflow, but all activities. For example, there is no need to return dependsOn in `def objects`.

realstraw:
(fixes #157) uses a separate workflow/dependency graph to manage dependency building

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
+++ b/core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
@@ -9,8 +9,9 @@ import scala.annotation.tailrec
 class WorkflowGraph private (
   val flow: Map[PipelineObjectId, Set[PipelineObjectId]],
   val activities: Map[PipelineObjectId, PipelineActivity]
-  // val roots: Set[PipelineObjectId]
 ) {
+
+  type Flow = Map[PipelineObjectId, Set[PipelineObjectId]]
 
   def this() =
     this(
@@ -83,10 +84,7 @@ class WorkflowGraph private (
     new WorkflowGraph(newFlow, newActivities)
   }
 
-  private def mergeFlow(
-    flow1: Map[PipelineObjectId, Set[PipelineObjectId]],
-    flow2: Map[PipelineObjectId, Set[PipelineObjectId]]
-  ): Map[PipelineObjectId, Set[PipelineObjectId]] =
+  private def mergeFlow(flow1: Flow, flow2: Flow): Flow =
     flow2.foldLeft(flow1) { case (f, (act, dependents)) =>
       val newDependents = f.get(act) match {
         case Some(ds) => ds ++ dependents
@@ -107,7 +105,10 @@ class WorkflowGraph private (
     } else {
       // get the immediate dependencies from the root node
       val rootDependents: Set[(PipelineObjectId, PipelineObjectId)] =
-        for { act <- (roots -- isolates); dependent <- flow(act) } yield (dependent, act)
+        for {
+          act <- (roots -- isolates)
+          dependent <- flow(act)
+        } yield (dependent, act)
 
       // assign dependees to the immediate dependents
       val actsWithDeps = rootDependents.groupBy(_._1)
--- a/core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
+++ b/core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
@@ -4,21 +4,52 @@ import com.krux.hyperion.activity.PipelineActivity
 import com.krux.hyperion.common.PipelineObjectId
 import scala.collection.mutable.Buffer
 import scala.language.implicitConversions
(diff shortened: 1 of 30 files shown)


core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
```diff
@@ -0,0 +1,123 @@
...
+import com.krux.hyperion.activity.PipelineActivity
+import com.krux.hyperion.common.PipelineObjectId
+import scala.collection.mutable.Buffer
+import scala.language.implicitConversions
+import scala.annotation.tailrec
+
+class WorkflowGraph private (
+  val flow: Map[PipelineObjectId, Set[PipelineObjectId]],
+  val activities: Map[PipelineObjectId, PipelineActivity]
+  // val roots: Set[PipelineObjectId]
```
sethyates: Remove dead code.

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
```diff
@@ -0,0 +1,123 @@
...
+) {
+
+  def this() =
```
sethyates: Why make the constructor above private and then have this?  Is this better than assigning default values to the constructor above?  Perhaps it is - just a Q.

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
```diff
@@ -0,0 +1,123 @@
...
+   * Merges two workflow graphs
+   */
+  def ++(other: WorkflowGraph): WorkflowGraph = {
+    val newFlow = mergeFlow(this.flow, other.flow)
+    val newActivities = this.activities ++ other.activities
+    new WorkflowGraph(newFlow, newActivities)
+  }
+
+  private def mergeFlow(
+    flow1: Map[PipelineObjectId, Set[PipelineObjectId]],
```
sethyates: Can't you define a `type` above for this Map type to clean up the code a bit?

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
realstraw: Oops, will do.

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
```diff
@@ -0,0 +1,123 @@
...
+  final def toActivities: Iterable[PipelineActivity] = {
+    assert(roots.size != 0)
+    assert(leaves.size != 0)
+
+    if (flow.size == 0) {
+      activities.values
+    } else {
+      // get the immediate dependencies from the root node
+      val rootDependents: Set[(PipelineObjectId, PipelineObjectId)] =
+        for { act <- (roots -- isolates); dependent <- flow(act) } yield (dependent, act)
```
sethyates: Nit: would be clearer for the iterators to be on a separate line, like so:
```
for {
  act <- (roots -- isolates)
  dependent <- flow(act)
} yield (dependent, act)
```

sethyates:
:+1: although I can't vouch for the correctness of the `WorkflowGraph` code just from a static inspection, so assuming you're confident the unit tests cover the core and edge cases, then LGTM :sparkles: :sparkles: :sparkles:

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
realstraw: Because I don't want other classes that uses workflow graph to set `flow` or `activities`, they are managed by `+`, `++`, and `~>`

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
realstraw: Good idea, will do

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
realstraw: OK, will do

core/src/main/scala/com/krux/hyperion/util/WorkflowGraph.scala
sethyates: Ah, strewth.  Duh.

PR #171, opened 2015-08-30
(Fixes #170) Output a count of pipeline objects

### LINKED ISSUE #170: Output a count of pipeline objects
By default, there is a limit of 100 objects in a pipeline.  We should output a count of pipeline objects so people can keep track of how close they are to the limit.

sethyates:


realstraw:
:+1:

PR #183, opened 2015-09-02
(build) remove travis-specific scala version cross compiling and just…

sethyates:
… use sbt's cross compile to avoid a race condition with pushing to Sonatype


realstraw:

PR #185, opened 2015-09-02
(Fixes #184) Properties for new notification activities are not prope…

### LINKED ISSUE #184: Properties for new notification activities are not properly exposed in the Activity definition

sethyates:
…rly exposed in the Activity definition

### DIFF
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/SendSnsMessageActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/SendSnsMessageActivity.scala
@@ -14,7 +14,12 @@ case class SendSnsMessageActivity private (
   scriptUri: Option[S3Uri],
   jarUri: String,
   mainClass: String,
-  arguments: Seq[String],
+  topicArn: String,
+  message: String,
+  subject: Option[String],
+  region: Option[String],
+  structuredMessage: Boolean,
+  attributes: Map[String, (String, String)],
   runsOn: Resource[Ec2Resource],
   dependsOn: Seq[PipelineActivity],
   preconditions: Seq[Precondition],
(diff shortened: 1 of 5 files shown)


realstraw:
:+1: :sparkles: :sparkles: :sparkles: LGTM

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/SendSnsMessageActivity.scala
```diff
@@ -31,6 +36,14 @@ case class SendSnsMessageActivity private (
   def named(name: String) = this.copy(id = PipelineObjectId.withName(name, id))
   def groupedBy(group: String) = this.copy(id = PipelineObjectId.withGroup(group, id))
 
+  def withSubject(subject: String) = this.copy(subject = Option(subject))
+  def withRegion(region: String) = this.copy(region = Option(region))
+  def withStructuredMessage = this.copy(structuredMessage = true)
+  def withAttribute(key: String, value: String, dataType: String = "String") = {
```
sethyates: Scala's parser is apparently brain damaged in the presence of something like:
```
 this.copy(attributes = attributes + (key, (dataType, value)))
```

Guess it can't tell if we're creating a tuple or enclosing an expression.

sethyates: And I didn't really want to do
```
new Tuple2(key, new Tuple2(dataType, value))
```

realstraw: Does `this.copy(attributes = attributes + ((key, (dataType, value))))` work?

sethyates: No

realstraw: How about `this.copy(attributes = attributes.+((key, (dataType, value))))`?

sethyates: Dunno didn't try not going to. That looks more hideous than what I have.

PR #187, opened 2015-09-02
(Fixes #186, #188, #189) Fixes to SFTP and File activities to support _SUCCESS and change collection constructors to use .empty

### LINKED ISSUE #186: LOW: Change collection constructors to use `.empty`
In many cases, we're using `Seq()` or `List()` or `Map()` where we should probably be using `Seq.empty` or `List.empty / Nil` or `Map.empty`.

sethyates:


ysalomon:
LGTM, my only concern, and i don't know the answer to this, is can we think of a case where we have valid files with only coupe of characters in them (like 1 row?) which i am guessing will still give us same size file?

sethyates:
I couldn't think of a valid reason for that, but open for counter points.  That was the most hacky part of this but seemed a pragmatic trade-off. :-(

realstraw:
:+1: LGTM

ysalomon:
:+1:

PR #194, opened 2015-09-03
(Fixes #191) Create a SparkActivity-type step that runs a single step…

### LINKED ISSUE #191: Create a SparkActivity-type step that runs a single step using HadoopActivity instead of MapReduceActivity

sethyates:
… using HadoopActivity instead of MapReduceActivity


realstraw:
:+1: we still need to test if works though.

PR #199, opened 2015-09-03
(fixes #160) better sns alarm format support

### LINKED ISSUE #160: Better/templated SNS alarm support
Current alarm message requires special syntax such as `#{node.@status}` and `#{node.name}`. I think we can provide standard or custom alarm templates which hides the details, and allows better and easier alarm.

realstraw:


core/src/main/scala/com/krux/hyperion/expression/RuntimeNode.scala
```diff
@@ -0,0 +1,65 @@
+package com.krux.hyperion.expression
+
+class RuntimeNode(runtimeSlot: RuntimeSlot) extends Expression {
```
sethyates: All other descendents of `Expression` end in `Exp`.  Not a problem but this does introduce an inconsistency.  The question is - does it matter?

sethyates:
:+1: LGTM :sparkles: :sparkles: one little nit on naming

PR #203, opened 2015-09-13
(Fixes #202) WorkflowGraph fails with assertion if not using named

### LINKED ISSUE #202: Failures when activities not using `named`
With the new `WorkflowGraph`, several pipelines that don't use `named` are failing to generate due to assertion.

sethyates:


sethyates:
For bonus points, I've also sorted `dependsOn` so unit tests are stable when comparing dependencies.

core/src/main/scala/com/krux/hyperion/common/PipelineObjectId.scala
```diff
@@ -20,32 +20,34 @@ object PipelineObjectId {
...
-case class NameGroupObjectId(name: String, group: String) extends PipelineObjectId {
+case class NameGroupObjectId(name: String, group: String, rand: String = UUID.randomUUID.toString) extends PipelineObjectId {
 
   val uniqueId = (name, group) match {
-    case ("", "") => UUID.randomUUID.toString
-    case ("", g) => s"${g}_${UUID.randomUUID.toString}"
-    case (n, "") => s"${n}_${UUID.randomUUID.toString}"
-    case (n, g) => s"${n}_${g}_${UUID.randomUUID.toString}"
+    case ("", "") => rand
+    case ("", g) => (g :: rand :: Nil).mkString("_")
```
realstraw: It's not much a difference but, isn't `List(g, rand).mkString("_")` a bit simpler? So you don't need the "Nil" at the end.

realstraw:

realstraw:
:rainbow: :fireworks: :sun_with_face: for bonus points :smiley:

core/src/main/scala/com/krux/hyperion/common/PipelineObjectId.scala
sethyates: But you need `List` at the beginning.  Much of a muchness.

PR #208, opened 2015-09-23
(fixes #207) Make workflow expression DSL avaible to pipeline def by …

### LINKED ISSUE #207: Make it not needed to import com.krux.hyperion.WorkflowExpression._
Since WorkflowExpression is the only supported way of defining pipelines, there is no need to make the implementing pipelines to always import `com.krux.hyperion.WorkflowExpression._`

realstraw:
…default

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/workflow/Implicits.scala
+++ b/core/src/main/scala/com/krux/hyperion/workflow/Implicits.scala
@@ -1,0 +1,20 @@
+package com.krux.hyperion.workflow
+
+import com.krux.hyperion.activity.PipelineActivity
+
+import scala.language.implicitConversions
+
+trait Implicits {
+
+  implicit def workflowIterable2WorkflowExpression(activities: Iterable[WorkflowExpression]): WorkflowExpression =
+    activities.reduceLeft(_ + _)
+
+  implicit def activityIterable2WorkflowExpression(activities: Iterable[PipelineActivity]): WorkflowExpression =
+    activities.map(activity2WorkflowExpression).reduceLeft(_ + _)
+
+  implicit def activity2WorkflowExpression(activity: PipelineActivity): WorkflowExpression =
+    WorkflowActivityExpression(activity)
+
+}
+
+object Implicits extends Implicits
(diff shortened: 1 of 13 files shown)


core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpression.scala
```diff
@@ -49,16 +46,3 @@ case class WorkflowActivityExpression(activity: PipelineActivity) extends Workfl
 case class WorkflowArrowExpression(left: WorkflowExpression, right: WorkflowExpression) extends WorkflowExpression
 
 case class WorkflowPlusExpression(left: WorkflowExpression, right: WorkflowExpression) extends WorkflowExpression
-
-object WorkflowExpression {
```
sethyates: I thought the default behaviour was to look in the companion object for implicit conversions...

core/src/main/scala/com/krux/hyperion/workflow/Implicits.scala
```diff
@@ -0,0 +1,20 @@
+package com.krux.hyperion.workflow
+
+import com.krux.hyperion.activity.PipelineActivity
+
+import scala.language.implicitConversions
```
sethyates: Nit: I would tend to put the scala imports first.

core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpression.scala
sethyates: Also, if you get rid of the companion object, does that break existing `import WorkflowExpression._` instances?  If so, this would be a breaking change and that's not good for a minor version bump.

realstraw: The default behavior that look into the companion object is if we have implicits that converts `WorkflowExpression` to some other type. This is not the case.

It does break the existing `import WorkflowExpression._` instances. I can make it backwards compatible by adding `object WorkflowExpression extends Implicits` if that is preferred.

core/src/main/scala/com/krux/hyperion/workflow/Implicits.scala
realstraw: Cool, I'll switch it around

core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpression.scala
realstraw: Ah, also I moved `WorkflowExpression` from `com.krux.hyperion` to `com.krux.hyperion.workflow` so it's a breaking change already. Should I revert this as well?

realstraw:
@sethyates how about this time? I've also renamed some stuff in this commit. Still a breaking change, if we want backwards compatibility, I can simply change the `WorkflowExpression` package back to `com.krux.hyperion`.

sethyates:
No breaking changes in minor releases. We can do the renames in 3.0.

realstraw:
@sethyates fixed, now no break

sethyates:
:+1: :sparkles: even though technically moving WorkflowGraph is a breaking change, nobody should be depending on it externally.

PR #221, opened 2015-10-12
(fixes #220) spark activity download jars to different locations

### LINKED ISSUE #220: Spark activity should download the jar to different files to avoid race condition of jobs running in parallel

realstraw:


sethyates:
:+1:

PR #235, opened 2015-10-27
(Fixes #234) SplitMergeFiles should allow ignoring cases where there is no input files

### LINKED ISSUE #234: SplitMergeFiles should allow ignoring cases where there is no input files

sethyates:


realstraw:
:+1: :rainbow: :railway_car: :rainbow:

PR #238, opened 2015-10-28
(Fixes #237) Allow Spark*Activity to override driver-memory

### LINKED ISSUE #237: Allow Spark*Activity to override driver-memory

sethyates:


core/src/main/scala/com/krux/hyperion/Implicits.scala
```diff
@@ -33,4 +33,16 @@ object Implicits {
   implicit def string2UniquePipelineId(prefix: String): PipelineObjectId = PipelineObjectId(prefix)
 
   implicit def stringContext2S3UriHelper(sc: StringContext): S3StringContext = S3StringContext(sc)
+
+  implicit class intToMemorySpec(n: Int) {
```
realstraw: We should use upper case for class and avoid using "Spec" as it's used for all unit test classes. Also we currently have the convention of `2` instead of `To` for implicit conversions borrowed from the scala codebase (although they are not consistent neither).

sethyates: Fixed

realstraw:
:+1: :rainbow: :rainbow: :rainbow:

PR #254, opened 2015-11-06
(Fixes #209) Override start activation time on command line

### LINKED ISSUE #209: Override start activation time on command line
Perhaps add a parameter to `--activate` or add a `--start` / `--today` that can specify `yyyy-mm-dd`, `now`, `today`, `yesterday` or `tomorrow`.

sethyates:


core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
```diff
@@ -28,11 +28,30 @@ class HyperionAwsClient(regionId: Option[String] = None, roleArn: Option[String]
     }
   }
 
+  /**
+    * Helper class to write a pipeline def and override its schedule.
+    *
+    * @param inner The inner pipeline to override.
+    * @param schedule The schedule to override with.
+    */
+  private case class ScheduleOverridingDataPipelineDef(
```
realstraw: I think you can create a DataPipeline case class that can be used to change it members. I'm going to create one for the parameter cli as well. So every time we change a member, we just create instance of the following case class.

```scala
case class AbstractDataPipeline(
  schedule: Schedule,
  workflow: WorkflowExpression,
  override val tags: Map[String, Option[String]],
  override val parameters: Iterable[Parameter[_]],
  override val pipelineName: String
) extends DataPipelineDef
```

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
```diff
@@ -28,11 +28,30 @@ class HyperionAwsClient(regionId: Option[String] = None, roleArn: Option[String]
...
+    override def objects = inner.objects
+    override def pipelineName = inner.pipelineName
+  }
+
   case class ForPipelineDef(
-    pipelineDef: DataPipelineDef,
-    customName: Option[String] = None
+    pipeline: DataPipelineDef,
+    customName: Option[String] = None,
+    schedule: Option[Schedule] = None
```
realstraw: I reckon `schedule` and `customName` should not be in AwsClient, and the aws client should use whatever is defined in pipeline. Since we now allow changing it by the case class now, it should be handled before it gets passed to the client.

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
sethyates: isn't that what I did?

realstraw: I mean as a separate class, not private in here.

core/src/main/scala/com/krux/hyperion/HyperionCli.scala
```diff
@@ -58,7 +88,21 @@ trait HyperionCli { this: DataPipelineDef =>
               case (k, v) => (k, Option(v))
             }
             c.copy(tags = c.tags + tag)
-          } unbounded()
+          } unbounded(),
+          opt[String]("start").action { (x, c) => c.copy(schedule = x.toLowerCase match {
```
realstraw: Not that different but alternatively you can have all this part implemented in
```
implicit val scheduleRead: scopt.Read[Schedule] = ...
```

then here you can do `opt[Schedule]("start")...`

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
sethyates: Yeah I agree.

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
```diff
@@ -52,38 +54,40 @@ class HyperionAwsClient(regionId: Option[String] = None, roleArn: Option[String]
...
+        // if using Hyperion for all DataPipeline management, this should never happen
+        case _ :: _ :: other => throw new Exception("Duplicated pipeline name")
+        case id :: Nil => Option(id)
         case Nil => None
       }
     }
 
-    def createPipeline(force: Boolean = false, tags: Map[String, Option[String]] = Map.empty): Option[String] = {
-      println(s"Creating pipeline $pipelineName")
+    def createPipeline(force: Boolean, activate: Boolean): Boolean = {
```
realstraw: Why change to `Boolean`? We do have (planned) use cases that the returned pipeline id is used, when using `HyperionAwsClient` directly in code rather than in CLI.

core/src/main/scala/com/krux/hyperion/HyperionCli.scala
```diff
@@ -1,41 +1,26 @@
...
+import java.io.File
 
-import com.krux.hyperion.workflow.WorkflowGraphRenderer
 import scopt.OptionParser
-import org.json4s.jackson.JsonMethods._
-import com.krux.hyperion.DataPipelineDef._
 
-trait HyperionCli { this: DataPipelineDef =>
+import com.github.nscala_time.time.Imports._
+import com.krux.hyperion.cli._
```
realstraw: :+1: :clap:

core/src/main/scala/com/krux/hyperion/cli/Options.scala
```diff
@@ -0,0 +1,40 @@
+package com.krux.hyperion.cli
+
+import java.io.File
+
+import com.krux.hyperion.Schedule
+
+trait Options {
```
realstraw: Why do we need this trait?

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
sethyates: Consistency with the other commands and it now activates directly. Wasn't aware of the other use cases.

core/src/main/scala/com/krux/hyperion/cli/Options.scala
sethyates: Mainly being lazy. Was trying to avoid a circular reference when defining Cli and Action when I had this all in one file. Probably moot now. I'll try to merge these two.

core/src/main/scala/com/krux/hyperion/HyperionAwsClient.scala
realstraw: Can return the created pipeline id without complicate things too much? 
Maybe add a new method without the `activate` flag? Something like:
```scala
def createPipeline(force: Boolean): Option[String]

def createPipeline(force: Boolean, activate: Boolean): Boolean
```

sethyates: Yeah that was the plan

realstraw:
:+1: :santa: :fireworks: :icecream:

PR #257, opened 2015-11-08
(Fixes #256) Use a logging framework instead of println - Commons Logging version

### LINKED ISSUE #256: Use a logging framework instead of println
To support the `HyperionAwsClient` in other applications outside of a CLI, we should make use of a logging framework.  Commons-Logging 1.1.3 is already included via AWS SDK, so perhaps we should just use that to avoid an additional dependency.

sethyates:


contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileRepartitioner.scala
```diff
@@ -85,7 +85,7 @@ case class FileRepartitioner(options: Options) {
 
     // Mark a successful job
     if (options.markSuccessfulJobs) {
-      Paths.get(System.getenv("OUTPUT1_STAGING_DIR"), "_SUCCESS").toFile.createNewFile()
+      Paths.get(dir.getPath, "_SUCCESS").toFile.createNewFile()
```
sethyates: This was actually a bug.

core/src/main/scala/com/krux/hyperion/cli/Action.scala
```diff
@@ -3,5 +3,5 @@ package com.krux.hyperion.cli
 import com.krux.hyperion.DataPipelineDef
 
 private[hyperion] trait Action {
-  def execute(options: Options, pipelineDef: DataPipelineDef): Boolean
+  def apply(options: Options, pipelineDef: DataPipelineDef): Boolean
```
sethyates: Realised after the fact that these appear as Function objects so should just use `apply` instead of `execute`.

build.sbt
```diff
@@ -11,6 +11,7 @@ val scoptArtifact           = "com.github.scopt"       %% "scopt"
 val jschArtifact            = "com.jcraft"             %  "jsch"                      % "0.1.53"
 val configArtifact          = "com.typesafe"           %  "config"                    % "1.2.1"
 val commonsIoArtifact       = "commons-io"             %  "commons-io"                % "2.4"
+val commonsLoggingArtifact  = "commons-logging"        % "commons-logging"            % "1.1.3"
```
realstraw: we already introduced slf4j in 3.0.0 should we use that one?

sethyates: Or should that one use commons logging?

realstraw: Most of the stuff I read says slf4j is better in terms of "SLF4J is a cleaner depen­dency and more effi­cient at run­time than commons-logging" and "Com­mons Log­gins does not pro­vide bind­ings to many com­mon log­ging frame­works". Cannot find too many articles vouch for commons logging.

sethyates: So you can just replace `commons-logging.jar` with `slf4j-over-jcl.jar` and commons logging is gone.  I've just tried replacing with straight SLF4J and without making any other differences and ran the ExampleSpark pipeline `create` and just get the following:
````
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
````

So the default case with SLF4J is NO-OP.  No logging.  Which seems pretty harsh.  Using this would require implementers to choose an SLF4J backend (such as LogBack).  The other case is to just use the Commons Logging interface which is already in the classpath via AWS SDK and if an implementer wants to swap out to use SLF4J they can swap out for `slf4j-over-jcl.jar` (which they probably do anyway).  Regardless, with Commons Logging, we get log output either way.

sethyates: Here is the SLF4J version: https://github.com/krux/hyperion/pull/260

realstraw:
:+1: this one as well, but still prefer https://github.com/krux/hyperion/pull/260 :-)

PR #260, opened 2015-11-08
(Fixes #256) Use a logging framework instead of println - SLF4J version

### LINKED ISSUE #256: Use a logging framework instead of println

sethyates:


realstraw:
:+1:  but I don't mind too much about `commons-logging` after seeing it gets pulled in by aws anyway. There are some threads in AWS that's trying to replace jcl with slf4j as well (but they need to make httpclient optional first). So I still prefer this one :-)

sethyates:
You do realise with this version that there won't be ANY logging unless the implementer chooses an SLF4J backend?  And being a library, we shouldn't be making that choice for them.

realstraw:
Yes, we can just put `slf4j-simple` or logback in the example project, so when we run the examples we don't get the warnings.

PR #265, opened 2015-11-09
(fixes #264) CLI schedule override only the explicitly specified part

### LINKED ISSUE #264: Overriding schedule in CLI behavior
Current implementation if specify `--start`, it will override the pipeline schedule with default interval (i.e. `1 day`) and default scheduling (i.e. `cron`). If the defined pipeline has an original interval of `1 month` this will be changed to `1 day` implicitly (`timeSeries` will be changed to `cron` as well), I think this is an unexpected behavior and the overriding scheduling properties should only override the part that is specified in the CLI explicitly.

And also we should be able to override `--every` and `--until` without `--start`. Currently they are silently ignored if `--start` is not present.

realstraw:


sethyates:
:+1: :santa: :fireworks: :shaved_ice:

PR #272, opened 2015-11-16
(fixes #271) Separate CLI with DataPipelineDef

### LINKED ISSUE #271: Separate CLI with DataPipelineDef
Make `DataPipelineDef` not mixing in `HyperionCli`, so application need to mixing in `HyperionCli` explicitly. This is required that sometimes traits need to be created, and having a companion object with Cli. The current implementation will trigger error:

```
ThePipeline has a main method with parameter type Array[String], but com.krux.ThePipeline will not be a runnable program.
```

realstraw:


sethyates:
:+1: :open_mouth: :cake:  :open_mouth: :cake:  :open_mouth: :cake:  :open_mouth: :cake:

PR #274, opened 2015-11-23
(fixes #273) the rest of Data Conversion Parameters to redshift copy …

### LINKED ISSUE #273: Add `ACCEPTINVCHARS` and the rest of Data Conversion Parameters to redshift copy options
Add `ACCEPTINVCHARS` to redshift copy options

realstraw:
…options

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/activity/RedshiftCopyOption.scala
+++ b/core/src/main/scala/com/krux/hyperion/activity/RedshiftCopyOption.scala
@@ -1,100 +1,53 @@
 package com.krux.hyperion.activity
 
-trait RedshiftCopyOption {
-  def repr: Seq[String]
-}
+case class RedshiftCopyOption(repr: Seq[String])
 
 object RedshiftCopyOption {
 
-  def csv(quote: String) = new RedshiftCopyOption {
-    def repr = Seq("CSV", "QUOTE", quote)
-  }
+  def csv(quote: String) = RedshiftCopyOption(Seq("CSV", "QUOTE", quote))
 
-  def csv = new RedshiftCopyOption {
-    def repr = Seq("CSV")
-  }
+  def csv = RedshiftCopyOption(Seq("CSV"))
 
-  def gzip = new RedshiftCopyOption {
-    def repr = Seq("GZIP")
-  }
+  def gzip = RedshiftCopyOption(Seq("GZIP"))
 
-  def delimiter(delChar: String) = new RedshiftCopyOption {
-    def repr = Seq("DELIMITER", s"'$delChar'")
-  }
+  def delimiter(delChar: String) = RedshiftCopyOption(Seq("DELIMITER", s"'$delChar'"))
 
-  def escape = new RedshiftCopyOption {
-    def repr = Seq("ESCAPE")
-  }
+  def escape = RedshiftCopyOption(Seq("ESCAPE"))
 
-  def nullAs(nullStr: String) = new RedshiftCopyOption {
-    def repr = Seq("NULL", s"'$nullStr'")
-  }
+  def nullAs(nullStr: String) = RedshiftCopyOption(Seq("NULL", s"'$nullStr'"))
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/krux/hyperion/activity/RedshiftCopyOption.scala
```diff
@@ -34,4 +34,67 @@ object RedshiftCopyOption {
     def repr = Seq("MAXERROR", errorCount.toString)
   }
 
+  def acceptInvChars = new RedshiftCopyOption {
```
sethyates: Rather than creating all these anonymous classes, I would feel slightly cleaner if we simply had a RedshiftCopyOptiom that took the repr Seq in the constructor

realstraw: Agreed, but I prefer not to do any refactoring work in 2.x

sethyates: Doesn't seem that disruptive to just do it now.  Are you envisaging other ways of implementing this?  I was just thinking:
```
case class RedshiftCopyOption(repr: Seq[String])

object RedshiftCopyOption {

  def csv(quote: String) = RedshiftCopyOption(Seq("CSV", "QUOTE", quote))

}
```

You could of course get fancier with a custom apply, etc.

realstraw: I actually didn't think or envisage anything ;-) I'll change it to the case class implementation

realstraw:
@sethyates ^^

core/src/main/scala/com/krux/hyperion/activity/RedshiftCopyOption.scala
```diff
@@ -1,37 +1,53 @@
...
-  def csv(quote: String) = new RedshiftCopyOption {
-    def repr = Seq("CSV", "QUOTE", quote)
-  }
+  def csv(quote: String) = RedshiftCopyOption(Seq("CSV", "QUOTE", quote))
+
+  def csv = RedshiftCopyOption(Seq("CSV"))
+
+  def gzip = RedshiftCopyOption(Seq("GZIP"))
+
+  def delimiter(delChar: String) = RedshiftCopyOption(Seq("DELIMITER", s"'$delChar'"))
```
sethyates: I know this is going further down the rabbit hole than you wanted, but the `s"'$something'" pattern is repeated several times here.  Would be nice to abstract as a:
```
private det quoted(s: String) = s"'$s'"
```
And then you could do
```
def delimiter(delChar: String) = RedshiftCopyOption(Seq("DELIMITER", quoted(delChar)))
```

sethyates:
:+1: :santa: only a very minor nit

realstraw:
@sethyates added quoted ^^

core/src/main/scala/com/krux/hyperion/activity/RedshiftCopyOption.scala
```diff
@@ -1,37 +1,56 @@
...
+case class RedshiftCopyOption(repr: Seq[String])
 
 object RedshiftCopyOption {
 
-  def csv(quote: String) = new RedshiftCopyOption {
-    def repr = Seq("CSV", "QUOTE", quote)
-  }
+  def quoted[T](x: T): String = s"'$x'"
+
+  def csv(quote: String) = RedshiftCopyOption(Seq("CSV", "QUOTE", quote))
```
sethyates: So this `quote` isn't, er, `quoted`?

sethyates:
LGTM :sparkles: :sparkles: :sparkles:

core/src/main/scala/com/krux/hyperion/activity/RedshiftCopyOption.scala
realstraw: Good catch, it should be

PR #276, opened 2015-12-11
(Fixes #275) Schedule is not honouring settings in non-application.co…

### LINKED ISSUE #275: Schedule is not honouring settings in non-application.conf config

sethyates:
…nf config

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/DataPipelineDefWrapper.scala
+++ b/core/src/main/scala/com/krux/hyperion/DataPipelineDefWrapper.scala
@@ -7,6 +7,7 @@ import com.krux.hyperion.parameter.Parameter
   * in order to override aspects.
   */
 private[hyperion] case class DataPipelineDefWrapper(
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/krux/hyperion/DataPipelineDefWrapper.scala
```diff
@@ -7,6 +7,7 @@ import com.krux.hyperion.parameter.Parameter
   * in order to override aspects.
   */
 private[hyperion] case class DataPipelineDefWrapper(
+  override val hc: HyperionContext,
```
realstraw: so this doesn't have to be implicit?

sethyates: No because nothing within `DataPipelineDefWrapper` needs it to be implicit.

realstraw:
:+1: :sparkles: :santa: :sparkles:

PR #279, opened 2015-12-15
(fixes #243) big refactor organize methods and fields and avoid duplicated code

### LINKED ISSUE #243: Refactor the pipeline objects so that the case class does not need to reimplement all the with* methods
There are a lot of copy paste for the pipeline objects e.g. `named`, `onSuccess`, `onFail` to name a few. Investigate ways to avoid the copy paste.

realstraw:
more can be done, but I think this is a good place to stop.


contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
```diff
@@ -83,21 +35,11 @@ object GoogleStorageDownloadActivity extends RunnableObject {
...
-      preconditions = Seq.empty,
-      onFailAlarms = Seq.empty,
-      onSuccessAlarms = Seq.empty,
-      onLateActionAlarms = Seq.empty,
-      attemptTimeout = None,
-      lateAfterTimeout = None,
-      maximumRetries = None,
-      retryDelay = None,
-      failureAndRerunMode = None
+      googleStorageUri = ""
```
sethyates: Should make this a required parameter to `apply`.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageUploadActivity.scala
```diff
@@ -2,101 +2,45 @@ package com.krux.hyperion.activity
...
-      preconditions = Seq.empty,
-      onFailAlarms = Seq.empty,
-      onSuccessAlarms = Seq.empty,
-      onLateActionAlarms = Seq.empty,
-      attemptTimeout = None,
-      lateAfterTimeout = None,
-      maximumRetries = None,
-      retryDelay = None,
-      failureAndRerunMode = None
+      googleStorageUri = ""
```
sethyates: Should make this a required parameter of `apply`.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PythonActivity.scala
```diff
@@ -2,103 +2,45 @@ package com.krux.hyperion.activity
...
-    attemptTimeout: Option[HDuration] = attemptTimeout,
-    lateAfterTimeout: Option[HDuration] = lateAfterTimeout,
-    maximumRetries: Option[HInt] = maximumRetries,
-    retryDelay: Option[HDuration] = retryDelay,
-    failureAndRerunMode: Option[FailureAndRerunMode] = failureAndRerunMode
-  ) = new PythonActivity(id, scriptUri, pythonScriptUri, pythonScript, pythonModule, pythonRequirements,
-    pipIndexUrl, pipExtraIndexUrls, arguments, input, output, stdout, stderr, runsOn, dependsOn, preconditions,
-    onFailAlarms, onSuccessAlarms, onLateActionAlarms, attemptTimeout, lateAfterTimeout, maximumRetries,
-    retryDelay, failureAndRerunMode)
+  override def withArguments(args: HString*) = copy(arguments = arguments ++ args)
```
sethyates: Why does this need to be overridden?

core/src/main/scala/com/krux/hyperion/activity/HadoopActivity.scala
```diff
@@ -16,58 +14,37 @@ import com.krux.hyperion.resource.{Resource, EmrCluster}
...
-  def onSuccess(alarms: SnsAlarm*) = this.copy(onSuccessAlarms = onSuccessAlarms ++ alarms)
-  def onLateAction(alarms: SnsAlarm*) = this.copy(onLateActionAlarms = onLateActionAlarms ++ alarms)
-  def withAttemptTimeout(timeout: HDuration) = this.copy(attemptTimeout = Option(timeout))
-  def withLateAfterTimeout(timeout: HDuration) = this.copy(lateAfterTimeout = Option(timeout))
-  def withMaximumRetries(retries: HInt) = this.copy(maximumRetries = Option(retries))
-  def withRetryDelay(delay: HDuration) = this.copy(retryDelay = Option(delay))
-  def withFailureAndRerunMode(mode: FailureAndRerunMode) = this.copy(failureAndRerunMode = Option(mode))
+  def withArguments(arguments: HString*) = copy(arguments = arguments ++ arguments)
+  def withHadoopQueue(queue: HString) = copy(hadoopQueue = Option(queue))
+  def withInput(input: S3DataNode*) = copy(inputs = inputs ++ input)
```
sethyates: `WithS3{In,Out}put` doesn't support this?

core/src/main/scala/com/krux/hyperion/activity/HiveCopyActivity.scala
```diff
@@ -84,26 +60,16 @@ case class HiveCopyActivity private (
...
 
 object HiveCopyActivity extends RunnableObject {
-  def apply(input: DataNode, output: DataNode)(runsOn: Resource[EmrCluster]): HiveCopyActivity =
+
+  def apply[A <: EmrCluster](input: DataNode, output: DataNode)(runsOn: Resource[A]): HiveCopyActivity[A] =
     new HiveCopyActivity(
-      id = PipelineObjectId(HiveCopyActivity.getClass),
+      baseFields = BaseFields(PipelineObjectId(HiveCopyActivity.getClass)),
+      activityFields = ActivityFields(runsOn),
+    emrTaskActivityFields = EmrTaskActivityFields(),
```
sethyates: Indentation.

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
```diff
@@ -1,91 +1,44 @@
...
-  def whenMet(conditions: Precondition*) = this.copy(preconditions = preconditions ++ conditions)
-  def onFail(alarms: SnsAlarm*) = this.copy(onFailAlarms = onFailAlarms ++ alarms)
-  def onSuccess(alarms: SnsAlarm*) = this.copy(onSuccessAlarms = onSuccessAlarms ++ alarms)
-  def onLateAction(alarms: SnsAlarm*) = this.copy(onLateActionAlarms = onLateActionAlarms ++ alarms)
-  def withAttemptTimeout(timeout: HDuration) = this.copy(attemptTimeout = Option(timeout))
-  def withLateAfterTimeout(timeout: HDuration) = this.copy(lateAfterTimeout = Option(timeout))
-  def withMaximumRetries(retries: HInt) = this.copy(maximumRetries = Option(retries))
-  def withRetryDelay(delay: HDuration) = this.copy(retryDelay = Option(delay))
-  def withFailureAndRerunMode(mode: FailureAndRerunMode) = this.copy(failureAndRerunMode = Option(mode))
+  def withOptions(opts: HString*) = super.withArguments(opts: _*)
```
sethyates: Really want to use `withArguments` here?  C&P error?

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
```diff
@@ -1,91 +1,44 @@
...
+  def options = shellCommandActivityFields.scriptArguments
```
sethyates: Really want to use `scriptArguments` here?  C&P error?

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
```diff
@@ -1,91 +1,44 @@
...
 
-  def objects: Iterable[PipelineObject] = runsOn.toSeq ++ input ++ output ++ dependsOn ++ preconditions ++ onFailAlarms ++ onSuccessAlarms ++ onLateActionAlarms
+  /**
+   * Arguments passed to the jar
+   */
+  override def withArguments(args: HString*) = this.copy(arguments = arguments ++ args)
```
sethyates: Removed `this` elsewwhere.

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
```diff
@@ -1,91 +1,44 @@
...
-  failureAndRerunMode: Option[FailureAndRerunMode]
-) extends PipelineActivity {
+  arguments: Seq[HString]
+) extends BaseShellCommandActivity with WithS3Input with WithS3Output {
 
-  def named(name: String) = this.copy(id = id.named(name))
-  def groupedBy(group: String) = this.copy(id = id.groupedBy(group))
+  type Self = JarActivity
+
+  assert(script.uri.nonEmpty)
```
sethyates: Why?

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
```diff
@@ -1,41 +1,29 @@
...
+import com.krux.hyperion.adt.HString
+import com.krux.hyperion.resource.{ Resource, SparkCluster }
 
 /**
  * Runs spark steps on given spark cluster with Amazon EMR
  */
 case class SparkActivity private (
-  id: PipelineObjectId,
+  baseFields: BaseFields,
+  activityFields: ActivityFields[SparkCluster],
```
sethyates: So why wouldn't you have `ActivityFields` extend `BaseFields`?

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
```diff
@@ -0,0 +1,107 @@
...
+ * Runs a Spark job on a cluster. The cluster can be an EMR cluster managed by AWS Data Pipeline
+ * or another resource if you use TaskRunner. Use SparkActivity when you want to run work in parallel.
+ * This allows you to use the scheduling resources of the YARN framework or the MapReduce resource
+ * negotiator in Hadoop 1. If you would like to run work sequentially using the Amazon EMR Step action,
+ * you can still use SparkActivity.
+ */
+case class SparkTaskActivity private (
+  baseFields: BaseFields,
+  activityFields: ActivityFields[SparkCluster],
+  emrTaskActivityFields: EmrTaskActivityFields,
```
sethyates: Likewise, why wouldn't you have `EmrTaskActivityFields` extend `ActivityFields`?

core/src/main/scala/com/krux/hyperion/activity/WithS3Input.scala
```diff
@@ -0,0 +1,20 @@
+package com.krux.hyperion.activity
+
+import com.krux.hyperion.adt.HBoolean
+import com.krux.hyperion.datanode.S3DataNode
+
+trait WithS3Input {
+
+  type Self <: WithS3Input
+
+  def shellCommandActivityFields: ShellCommandActivityFields
```
sethyates: Oh...  This is a rather odd coupling.

core/src/main/scala/com/krux/hyperion/common/NamedPipelineObject.scala
```diff
@@ -0,0 +1,18 @@
+package com.krux.hyperion.common
+
+trait NamedPipelineObject extends PipelineObject {
```
sethyates: Seems like rather a lot for just a single `id` field, doesn't it?  Do you envisage more going into the `BaseFields`?

core/src/main/scala/com/krux/hyperion/common/PipelineObject.scala
```diff
@@ -23,6 +28,6 @@ trait PipelineObject extends Ordered[PipelineObject] {
     }
   }
 
-  def compare(that: PipelineObject): Int =  this.id.compare(that.id)
+  def compare(that: PipelineObject): Int =  self.id.compare(that.id)
```
sethyates: Why is `self` important here?  Will the `def id` above not polymorphically pick up the right one?

core/src/main/scala/com/krux/hyperion/database/Database.scala
```diff
@@ -1,12 +1,33 @@
...
+  def databaseFields: DatabaseFields
+  def updateDatabaseFields(fields: DatabaseFields): Self
+
+  def username: HString = databaseFields.username
+  def withUserName(username: HString) = updateDatabaseFields(
+    databaseFields.copy(username = username)
+  )
+
+  def `*password` = databaseFields.`*password`
+  def withPassword(pass: HString) = updateDatabaseFields(
```
sethyates: `pass` -> `password`

core/src/main/scala/com/krux/hyperion/database/Database.scala
```diff
@@ -1,12 +1,33 @@
...
+    databaseFields.copy(`*password` = pass)
+  )
+
+  def databaseName = databaseFields.databaseName
+  def withDatabaseName(dbname: HString) = updateDatabaseFields(
```
sethyates: `dbname` -> `databaseName`

core/src/main/scala/com/krux/hyperion/database/JdbcDatabase.scala
```diff
@@ -1,39 +1,57 @@
...
+import com.krux.hyperion.adt.HString
 import com.krux.hyperion.aws.AdpJdbcDatabase
-import com.krux.hyperion.common.PipelineObjectId
+import com.krux.hyperion.common.{ PipelineObjectId, BaseFields }
 
 /**
  * Defines a JDBC database
  */
-trait JdbcDatabase extends Database {
+case class JdbcDatabase private (
```
sethyates: Yay!  Always disliked the "implement a trait as an object" false dichotomy for Database objects.

core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
```diff
@@ -84,26 +64,25 @@ case class Ec2Resource private (
...
-    securityGroups = Seq(hc.ec2SecurityGroup),
-    securityGroupIds = Seq.empty,
-    spotBidPrice = None,
-    useOnDemandOnLastAttempt = None,
-    initTimeout = None,
-    terminateAfter = hc.ec2TerminateAfter.map(duration2HDuration),
-    actionOnResourceFailure = None,
-    actionOnTaskFailure = None,
-    httpProxy = None
+    terminateAfter = hc.ec2TerminateAfter.map(x => x: HDuration)
```
sethyates: Guess `.map(_: HDuration)` doesn't work?

sethyates:
TL;DR - I commented on what I thought where the most salient points.  Overall like the direction.

core/src/main/scala/com/krux/hyperion/activity/HadoopActivity.scala
realstraw: It doesn't with the current implementation. I'll make some changes to support this.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
realstraw: Will do

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PythonActivity.scala
realstraw: Otherwise, the actual python command arguments will be added to the `scriptArguments`. Which needs to have the python script location and other python environment stuff such as the pip files etc. We can, however, not override this but make scriptArguments always start with the actual python related arguments, but this one seems to be arguably cleaner.

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
realstraw: No, the withOptions here will parse options to the `run-jar.sh` script (although it does not seem to take any). `withArguments` actually takes the argument passed to the `main`. This is to make this behave like it used to be, although I cannot remember why we have `withOptions`, maybe to support setting some jvm parameters?

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
realstraw: To stop people accidentally pass "s3://the/jar/location.jar" as a string instead of a script.

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
realstraw: because I cannot extend a case class.

core/src/main/scala/com/krux/hyperion/common/NamedPipelineObject.scala
realstraw: I'm planning to add `name` here when we start to make `name` not `id`.

core/src/main/scala/com/krux/hyperion/common/PipelineObject.scala
realstraw: I think this was for shapeless implementation, not needed anymore will remove.

core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
realstraw: It doesn't, that was my first thought as well ;-)

core/src/main/scala/com/krux/hyperion/activity/WithS3Input.scala
realstraw: Yes :-) this is why `WithS3{In,Out}put` doesn't support the input/output in other activities.

core/src/main/scala/com/krux/hyperion/activity/HadoopActivity.scala
realstraw: Actually, can I address this in a future PR? The main purpose of `WithS3{In,Out}put` as it is intended now is to disable setting input/output fields and enable custom `with{In,Out}put` methods for the contrib activities, and also set the `stage` field properly when io field is given.

The broader problem is that all `activity` have input/output data node with different requirements, and this looks like should be at activity level. But the current implementation of `AdpActivities` for these ones are not consistent, some requires Option[DataNode] some requires Seq[DataNode], and activities themselves have different contraints like only S3DataNode, or RedshiftDataNode. Don't want to dig too many rabbit holes in the same PR.

sethyates: Yeah of course.  Was merely a query.

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
sethyates: Does `BaseFields` have to be a case class?  The other reason I could think of would be the mutation of `id` would possibly be more complicated if not just updating `BaseFields`.

core/src/main/scala/com/krux/hyperion/common/NamedPipelineObject.scala
sethyates: Ack.

core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
sethyates: Yeah figured. :-(

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
```diff
@@ -1,91 +1,44 @@
...
-    onFail = seqToOption(onFailAlarms)(_.ref),
-    onSuccess = seqToOption(onSuccessAlarms)(_.ref),
-    onLateAction = seqToOption(onLateActionAlarms)(_.ref),
-    attemptTimeout = attemptTimeout.map(_.serialize),
-    lateAfterTimeout = lateAfterTimeout.map(_.serialize),
-    maximumRetries = maximumRetries.map(_.serialize),
-    retryDelay = retryDelay.map(_.serialize),
-    failureAndRerunMode = failureAndRerunMode.map(_.serialize)
-  )
+  def withOptions(opts: HString*) = super.withArguments(opts: _*)
```
sethyates: OK well this is a very non-obvious way to implement this.  Why wouldn't you just let `withArguments` (`scriptArguments`) be the arguments and add an `options` instead of `arguments`?

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
realstraw: yeah, we could change these fields classes to be normal classes with `var`s. Then these classes can extend each other. And would avoid some of the problems we had because of the immutable fields. I was actually tempted to change all fields to `var`s at some stage, and it will avoid all these repeated `def update*Fields`, but was afraid of that we might rely on assumption of immutable data somewhere.

core/src/main/scala/com/krux/hyperion/activity/JarActivity.scala
realstraw: ha, good point, will change

realstraw: fixed in https://github.com/realstraw/hyperion/commit/a1182f3092308b776638d1f251d1d0e933549e2b

realstraw:
closed by https://github.com/krux/hyperion/commit/0236d89fa8a49a20a8e355c306f3aea189b3eeab

PR #283, opened 2015-12-22
(fixes #282) add support for getting hyperion aws client by pipeline …

### LINKED ISSUE #282: In addition to ForPipelineId and ForPipelineDef, HyperionAwsClient also need to support ForName

realstraw:
…name


ysalomon:
:+1:  i like protected.

PR #288, opened 2015-12-28
(fixes #285) snsTopicArn is required and add default subject and message

### LINKED ISSUE #285: SnsAlarm need to be supplied with default subject and message
and make sure TopicArn is a required field to be supplied during object initialization.

realstraw:


ysalomon:
LGTM

PR #293, opened 2016-01-04
(fixes #284) clean up the implicits

### LINKED ISSUE #284: Clean up the implicits
to the the corresponding companion objects, implicit class or proper packages.

realstraw:


sethyates:
:+1: as long as you have sample workflows using the explicits that still compile.

PR #296, opened 2016-01-05
(fixes #248, fixes #295) parameter refactor, adding encrypted parameter type

### LINKED ISSUE #248: Enforce the encrypted-ness via type such that .encrypted returns EncryptedParameter[T]

### LINKED ISSUE #295: Parameter can be better refactored with adhoc polymorphism instead of reflection with TypeTags
It not only eliminate the use of reflection, but also have much better compile time type safety and better extensibility.

realstraw:


core/src/main/scala/com/krux/hyperion/expression/GenericParameter.scala
```diff
@@ -0,0 +1,131 @@
...
+
+import org.joda.time.{ DateTime, DateTimeZone }
+
+import com.krux.hyperion.common.S3Uri
+
+trait GenericParameter[T] {
+
+  type Exp <: TypedExpression
+
+  def parseString: (String) => T
```
sethyates: Why do it this way rather than `def parseString(s: String): T`?

core/src/main/scala/com/krux/hyperion/expression/GenericParameter.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  def ref(param: Parameter[T]): Exp
+
+  def `type`: String
+
+}
+
+object GenericParameter {
+
+  implicit object IntGenericParameter extends GenericParameter[Int] {
```
sethyates: I don't understand how these work, just from reading the code, so a big comment is probably in order here.

core/src/main/scala/com/krux/hyperion/expression/GenericParameter.scala
```diff
@@ -0,0 +1,131 @@
...
+
+    val parseString = (stringValue: String) => Duration(stringValue)
+
+    def ref(param: Parameter[Duration]): Exp = new Exp with Evaluatable[Duration] {
+      def content = param.name
+      def evaluate() = param.evaluate()
+    }
+
+    // this is not typo, the aws type of this Duration is String
+    def `type`: String = "String"
```
sethyates: Instead of a String return here then perhaps an enum with the correct `toString` would be clearer?  And you can put a big comment on the enum definition noting the relationship to AWS.  Or have a map that maps Enum -> AWS String.

sethyates:
I think we've finally exceeded the limit of the ability for the code to speak for itself.  The level of genericity and indirection here makes it pretty incomprehensible what is actually going on.  So, I think we either need to start investing in some comments or rethink how complex we want to make this.

core/src/main/scala/com/krux/hyperion/expression/GenericParameter.scala
realstraw: We actually need to have this `parseString` stored as a function at the time we lose the type. When we put the parameter in the parameter list as `List[Parameter[_]]`. Since we implemented compile time check, and we lose the type information in the `List`, we need the underlying parameter to know at the time it's constructed how to parse a `String` to it's parameterized type.

sethyates: Comment please.  Not immediately obvious why (obviously :-)).

core/src/main/scala/com/krux/hyperion/expression/GenericParameter.scala
realstraw: This is mainly the concept what people call adhoc-polymorphism in contrast to subtype-polymorphism. The most widely used place is in  https://github.com/scala/scala/blob/v2.11.7/src/library/scala/math/Numeric.scala and https://github.com/scala/scala/blob/v2.11.7/src/library/scala/math/Ordering.scala where you can do numeric operation on `T` as long as you have a implicit `Numeric[T]` in scope, or you can sort `T` as long as you have a implicit `Ordering[T]` is in scope.

core/src/main/scala/com/krux/hyperion/expression/GenericParameter.scala
realstraw: Good point, will do.

realstraw:
Will add docs with some description.

ysalomon:
maybe try not to mention polymorphism too much.

realstraw:
@sethyates added comments and changed the `type` of parameter to return Enum instead of String

ysalomon:
happy to :+1: for now to unblock job server release. Better documentation will be an ongoing effort that must be addressed.

realstraw:
merged

PR #297, opened 2016-01-09
3.0

realstraw:
A simple migration guide that documents incompatibility is in https://github.com/krux/hyperion/wiki/Hyperion-3.0


sethyates:
I've spot-checked this, but as you've noted elsewhere, we've already reviewed each individual commit, so this should be good to go. :sparkles: :sparkles: :sparkles:

PR #301, opened 2016-01-09
(fixes #300) Added more options to Parameter.encrypted and Parameter.…

### LINKED ISSUE #300: Parameter.encrypted is missing the (id, value) option
`apply` has both id, and (id, value). `encrypted` and `unencrypted` should contain similar options.

realstraw:
…unencrpted


sethyates:
LGTM

PR #302, opened 2016-01-09
(fixes #299) fix constant expression implicit to avoid unnecessary im…

### LINKED ISSUE #299: ConstantExpression's implicits need to be spread to individual Expression types rather than all one companion object

realstraw:
…port


sethyates:
LGTM

PR #303, opened 2016-01-09
(fixes #298) Make sequence of HType conversion implicitly available

### LINKED ISSUE #298: Need implicit conversions from Seq[String] => Seq[HString] and other hyperion adt collections

realstraw:


sethyates:
~LGTM

PR #309, opened 2016-01-15
(fixes #213) start use `name` field instead rely on `id`

### LINKED ISSUE #213: Start to use the `name` field, since `id` can be very restrictive
The initial design makes `name` and `id` for an object the same. We should make `named` and `groupedBy` update `name`, but not `id` since `name` is less restrictive and actually makes more sense.

realstraw:


sethyates:
:+1:

PR #311, opened 2016-01-20
(fixes #310) fix a bug preconditions does not include referenced objects

### LINKED ISSUE #310: Precondition is missing the referenced objects
need to include the referenced objects in `def objects`

realstraw:


ysalomon:
:+1:

PR #336, opened 2016-02-10
(Fixes #335) WorkflowGraphRenderer use name instead of id

### LINKED ISSUE #335: WorkflowGraphRenderer use name instead of id
Now we start to use `name` field. We need to update the `WorkflowGraphRenderer` to use the `name` field instead of `id`

sethyates:


realstraw:

PR #339, opened 2016-02-15
Adding a new Copy S3 Path Activity as a wrapper for aws s3 cp

roopak:


build.sbt
```diff
@@ -1,4 +1,4 @@
-val hyperionVersion = "3.2.0"
+val hyperionVersion = "3.2.1"
```
sethyates: 3.3.0

sethyates: Also raise an issue and add to CHANGELOG.md

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/CopyS3PathActivity.scala
```diff
@@ -0,0 +1,39 @@
+package com.krux.hyperion.activity
+
+import com.krux.hyperion.adt.HS3Uri
+import com.krux.hyperion.common.{PipelineObjectId, BaseFields}
+import com.krux.hyperion.expression.RunnableObject
```
sethyates: Format we use is space after { and before }

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/CopyS3PathActivity.scala
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Activity to Copy recursively between S3 Paths.
+ * This does not fail in case sourcePath does not exist
+ */
+case class CopyS3PathActivity private (
+  baseFields: BaseFields,
+  activityFields: ActivityFields[Ec2Resource],
+  shellCommandActivityFields: ShellCommandActivityFields,
+  sourceS3Path: HS3Uri,
```
sethyates: Would be more familiar to use withInput and withOutput overriding the operations from ShellCommandActivity (stage false).

sethyates: Although...that probably wouldn't work with DataNodes would it?

sethyates:
This belongs in contrib, not core.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/CopyS3PathActivity.scala
realstraw: It can still use DataNodes, but only the reference part not formats etc. It probably makes more sense to keep it as `HS3Uri` to be consistent with `DeleteS3PathActivity`.

realstraw:
In additional to @sethyates comments about raising an issue etc. All our pull request should follow the general github approach, that is, the pull request should be from your own fork of the repo rather than making new branches and put in this repo.

sethyates:
Would be worth adding acl and grants per SetS3AclActivity

realstraw:
As discussed with @roopak a few days ago, we are going to close this for now.

PR #350, opened 2016-03-17
(Fixes #349) SplitMergeFilesActivity needs to pass `temporary-directory`

### LINKED ISSUE #349: SplitMergeFilesActivity needs to pass `temporary-directory`

sethyates:


core/src/main/scala/com/krux/hyperion/adt/HType.scala
```diff
@@ -113,7 +113,12 @@ case class HDouble(value: Either[Double, DoubleExp]) extends HType with Optional
 
 }
 
-case class HBoolean(value: Either[Boolean, BooleanExp]) extends HType
+case class HBoolean(value: Either[Boolean, BooleanExp]) extends HType {
+  def exists[B](fn: => B) = value match {
```
realstraw: why call by name?

sethyates: huh?

realstraw: I mean why use `fn: => B` instead of `fn: B`?

sethyates: Isn't that typical for these sorts?  I don't want to have `fn` evaluated unless it's actually required.  My understanding is if I passed by value, Scala would evaluate the expression before passing into the function.

realstraw: I seems not matter in this case. Just curious, I thought there's a particular reason.

realstraw:

PR #352, opened 2016-03-25
(Fixes #351) Add ifExists to DeleteS3PathActivity

### LINKED ISSUE #351: Add whenExists to DeleteS3PathActivity
When specidified, perform delete only if the path exists, and if not exist, performs a no-op

arijitdey:
Adding ifExists to the DeleteS3PathActivity

### DIFF
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/DeleteS3PathActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/DeleteS3PathActivity.scala
@@ -21,6 +21,16 @@ case class DeleteS3PathActivity private (
   def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
   def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
 
+  def ifExists(existsS3Path: HS3Uri) = updateShellCommandActivityFields(
+    shellCommandActivityFields.copy(script =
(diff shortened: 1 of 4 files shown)


contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/DeleteS3PathActivity.scala
```diff
@@ -21,6 +21,16 @@ case class DeleteS3PathActivity private (
...
   def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
 
+  def ifExists(s3Path: HS3Uri) = updateShellCommandActivityFields(
+    shellCommandActivityFields.copy(script =
+        """count=`aws s3 ls %s | wc -l`;
+           if [[ $count -gt 0 ]];
+           then
+              aws s3 rm --recursive %s ;
+            fi
+        """.format(s3Path, this.s3Path).stripMargin
```
realstraw: can you use string interpolation instead of format.

e.g. `s"""count = `aws s3 ls $s3Path..."""`

Also `stripMargin` works only when you have `|` in the string.

e.g. 
```
"""abcd
   |efg
""".stripMargin
```
will evaluate to
```
abcd
efg
```
instead of
```
abcd
   efg
```

realstraw:
Please also update the build number in `build.sbt` and `CHANGE_LOG.md`

sethyates:
:+1: that would work

realstraw:
:+1: can you also stash your commits

PR #359, opened 2016-04-11
(Fixes #357) Adding recursive option to the GoogleStorageDowloadActivity

### LINKED ISSUE #357: Add recursive option to GoogleStorageDownloadActivity
Currently GoogleStorageDownloadActivity supports copying of strictly folders and not specific file. Add support for recursive option so that consumers can download both files and folders.

arijitdey:
Currently GoogleStorageDownloadActivity supports copying of strictly folders and not specific file. Adding support for recursive option so that consumers can download both files and folders.

### DIFF
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
@@ -1,10 +1,10 @@
 package com.krux.hyperion.activity
 
 import com.krux.hyperion.HyperionContext
-import com.krux.hyperion.adt.{ HS3Uri, HString }
-import com.krux.hyperion.common.{ BaseFields, PipelineObjectId }
+import com.krux.hyperion.adt.{HBoolean, HS3Uri, HString}
+import com.krux.hyperion.common.{BaseFields, PipelineObjectId}
 import com.krux.hyperion.expression.RunnableObject
-import com.krux.hyperion.resource.{ Ec2Resource, Resource }
+import com.krux.hyperion.resource.{Ec2Resource, Resource}
 
 /**
  * Google Storage Download activity
@@ -14,7 +14,8 @@ case class GoogleStorageDownloadActivity private (
   activityFields: ActivityFields[Ec2Resource],
   shellCommandActivityFields: ShellCommandActivityFields,
   botoConfigUrl: HS3Uri,
-  googleStorageUri: HString
+  googleStorageUri: HString,
+  ignoreMissingSource: HBoolean
 ) extends GoogleStorageActivity with WithS3Output {
 
   type Self = GoogleStorageDownloadActivity
@@ -22,6 +23,14 @@ case class GoogleStorageDownloadActivity private (
   def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
   def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
   def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def ifExists = copy(ignoreMissingSource = true)
+
+  override def scriptArguments = Seq(
+    botoConfigUrl.serialize: HString,
+    googleStorageUri,
+    ignoreMissingSource.serialize
+  )
 
 }
 
@@ -33,7 +42,8 @@ object GoogleStorageDownloadActivity extends RunnableObject {
       activityFields = ActivityFields(runsOn),
       shellCommandActivityFields = ShellCommandActivityFields(GoogleStorageActivity.downloadScript),
       botoConfigUrl = botoConfigUrl,
-      googleStorageUri = input
+      googleStorageUri = input,
+      ignoreMissingSource = HBoolean.False
     )
 
 }
--- a/scripts/activities/gsutil-download.sh
+++ b/scripts/activities/gsutil-download.sh
@@ -1,7 +1,7 @@
 #! /usr/bin/env bash
 
 usage() {
-  echo "usage: gsutil_download.sh s3://bucket/boto.config gs://gsbucket/path/"
+  echo "usage: gsutil_download.sh s3://bucket/boto.config gs://gsbucket/path/ true|false"
   echo
   echo "The contents of the GS path will be downloaded to OUTPUT1_STAGING_DIR (${OUTPUT1_STAGING_DIR})"
   exit 3
@@ -54,6 +54,9 @@ BOTO_CONFIG_SOURCE="$1"
 # This is the destination Google Storage location
 INPUT_GOOGLE_STORAGE="$2"
 
+# This is the flag to ignore if source uri is missing or throw an error.
+IGNORE_IF_SRC_MISSING="${3:-false}"
+
 # Download and extract the tarball.
 # We use --no-check-certificate because Google are naughty with their certificates.
 wget --no-verbose --no-check-certificate ${GSUTIL_URL}
@@ -62,5 +65,12 @@ tar -xzf ${TARBALL}
 # Download the boto configuration
 aws s3 cp ${BOTO_CONFIG_SOURCE} ${BOTO_CONFIG}
 
-./gsutil/gsutil cp ${INPUT_GOOGLE_STORAGE}/* ${OUTPUT1_STAGING_DIR}/
+COMMAND="./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/"
 
+if [[ ${IGNORE_IF_SRC_MISSING} == "true" ]]; then
+  if ! ${COMMAND} ; then
+    echo "INFO: no data available to download"
+  fi
+else
+  ${COMMAND}
+fi
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,13 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 3.2.5 - 2016-04-11
+### Added
+- [#357](https://github.com/krux/hyperion/issues/357) - Add recursive option to GoogleStorageDownloadActivity
+
 ## 3.2.4 - 2016-03-29
 ### Added
-- [#321](https://github.com/krux/hyperion/issues/321) - Add overloaded methods accepting HS3Uri for Activities 
+- [#321](https://github.com/krux/hyperion/issues/321) - Add overloaded methods accepting HS3Uri for Activities
 
 ## 3.2.3 - 2016-03-28
 ### Added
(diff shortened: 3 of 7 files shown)


scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +62,6 @@ tar -xzf ${TARBALL}
 # Download the boto configuration
 aws s3 cp ${BOTO_CONFIG_SOURCE} ${BOTO_CONFIG}
 
-./gsutil/gsutil cp ${INPUT_GOOGLE_STORAGE}/* ${OUTPUT1_STAGING_DIR}/
-
+if ! ./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/ ; then
```
sethyates: You have `bash` expressions available here, so I'd prefer to use that syntax if possible (like the rest of the file does).

realstraw:
Shouldn't we add recursive an option rather than changing the default? also ignore empty source should also be an option rather than a default behavior.

arijitdey:
@realstraw The default is recursive only as we append only `/*`. It's just using the gsutil's command line option instead of appending `/*`.

scripts/activities/gsutil-download.sh
arijitdey: @sethyates As tested bash expressions are not supported in this case.

realstraw:
@arijitdey cool, what about the ignore empty source?

arijitdey:
@realstraw We do have a friendly message.

realstraw:
So before it fails the activity, not it succeeds with a log message. It's a breaking change, and I think users should be able to specify different behaviors, and default to fail the activity so that it's not a breaking change.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
```diff
@@ -14,14 +14,19 @@ case class GoogleStorageDownloadActivity private (
...
+  googleStorageUri: HString,
+  ignoreMissingSource: HString
 ) extends GoogleStorageActivity with WithS3Output {
 
   type Self = GoogleStorageDownloadActivity
 
   def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
   def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
   def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+  def updateIgnoreMissingSource(fields: HString) = copy(ignoreMissingSource = fields)
```
realstraw: you don't need this, just make `def ifExists` do the copy.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
```diff
@@ -14,14 +14,19 @@ case class GoogleStorageDownloadActivity private (
   activityFields: ActivityFields[Ec2Resource],
   shellCommandActivityFields: ShellCommandActivityFields,
   botoConfigUrl: HS3Uri,
-  googleStorageUri: HString
+  googleStorageUri: HString,
+  ignoreMissingSource: HString
```
realstraw: Why not boolean?

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
```diff
@@ -23,6 +24,14 @@ case class GoogleStorageDownloadActivity private (
   def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
   def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
 
+  def ifExists = copy(ignoreMissingSource = true)
+
+  override def scriptArguments = Seq(
+    botoConfigUrl.serialize: HString,
+    googleStorageUri,
+    ignoreMissingSource.serialize
```
sethyates: Why do you need `.serialize` here?

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/GoogleStorageDownloadActivity.scala
```diff
@@ -33,7 +42,8 @@ object GoogleStorageDownloadActivity extends RunnableObject {
       activityFields = ActivityFields(runsOn),
       shellCommandActivityFields = ShellCommandActivityFields(GoogleStorageActivity.downloadScript),
       botoConfigUrl = botoConfigUrl,
-      googleStorageUri = input
+      googleStorageUri = input,
+      ignoreMissingSource = false
```
sethyates: `HBoolean.False`

scripts/activities/gsutil-download.sh
```diff
@@ -1,13 +1,13 @@
...
 usage() {
-  echo "usage: gsutil_download.sh s3://bucket/boto.config gs://gsbucket/path/"
+  echo "usage: gsutil_download.sh s3://bucket/boto.config gs://gsbucket/path/ true|false"
   echo
   echo "The contents of the GS path will be downloaded to OUTPUT1_STAGING_DIR (${OUTPUT1_STAGING_DIR})"
   exit 3
 }
 
-if ([ -z "$1" ] || [ -z "$2" ]); then
+if ([ -z "$1" ] || [ -z "$2" ] || [ -z "$3" ]); then
```
sethyates: This makes this a breaking change.  You should just default the third argument.

scripts/activities/gsutil-download.sh
```diff
@@ -54,6 +54,9 @@ BOTO_CONFIG_SOURCE="$1"
 # This is the destination Google Storage location
 INPUT_GOOGLE_STORAGE="$2"
 
+# This is the flag to ignore if source uri is missing or throw an error.
+IGNORE_IF_SRC_MISSING = "$3"
```
sethyates: `${3:-false}`

scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +65,10 @@ tar -xzf ${TARBALL}
 # Download the boto configuration
 aws s3 cp ${BOTO_CONFIG_SOURCE} ${BOTO_CONFIG}
 
-./gsutil/gsutil cp ${INPUT_GOOGLE_STORAGE}/* ${OUTPUT1_STAGING_DIR}/
-
+if [ ! -z "$3" -a "$3" == "true" ]; then
```
sethyates: `$3` at this point cannot be empty so the `-z "$3" -a` bit is redundant.

scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +65,10 @@ tar -xzf ${TARBALL}
...
+  if ! ./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/ ; then
+    echo "INFO: no data available to download"
+  fi
+else
+    ./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/
```
sethyates: Rather than duplicate, just capture the exit code from the utility and use for comparison.

scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +65,14 @@ tar -xzf ${TARBALL}
...
+./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/
+
+status=$?
 
+if [[ "$3" == "true" ]]; then
+  if [[ $status -ne 0 ]]; then
+    echo "INFO: no data available to download"
+  fi
+else
+  exit 3
```
realstraw: why `exit 3` instead of `exit $?`?

scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +65,14 @@ tar -xzf ${TARBALL}
 # Download the boto configuration
 aws s3 cp ${BOTO_CONFIG_SOURCE} ${BOTO_CONFIG}
 
-./gsutil/gsutil cp ${INPUT_GOOGLE_STORAGE}/* ${OUTPUT1_STAGING_DIR}/
+./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/
+
+status=$?
```
realstraw: we are using all caps for variables e.g. `IGNORE_IF_SRC_MISSING`, we should be consistent here as well

scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +65,14 @@ tar -xzf ${TARBALL}
 # Download the boto configuration
 aws s3 cp ${BOTO_CONFIG_SOURCE} ${BOTO_CONFIG}
 
-./gsutil/gsutil cp ${INPUT_GOOGLE_STORAGE}/* ${OUTPUT1_STAGING_DIR}/
+./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/
```
realstraw: we are using `-xe` doesn't that going to stop here and exit with non zero regardless?

scripts/activities/gsutil-download.sh
```diff
@@ -62,5 +65,12 @@ tar -xzf ${TARBALL}
...
 
-./gsutil/gsutil cp ${INPUT_GOOGLE_STORAGE}/* ${OUTPUT1_STAGING_DIR}/
+COMMAND="./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/"
 
+if [[ ${IGNORE_IF_SRC_MISSING} == "true" ]]; then
+  if ! ${COMMAND} ; then
+    echo "INFO: no data available to download"
+  fi
+else
+  ${COMMAND}
```
realstraw: Don't think this works. You might want to use `eval`

scripts/activities/gsutil-download.sh
```diff
@@ -65,14 +65,12 @@ tar -xzf ${TARBALL}
...
 aws s3 cp ${BOTO_CONFIG_SOURCE} ${BOTO_CONFIG}
 
-./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/
+COMMAND="./gsutil/gsutil -m cp -r ${INPUT_GOOGLE_STORAGE} ${OUTPUT1_STAGING_DIR}/"
 
-status=$?
-
-if [[ "$3" == "true" ]]; then
-  if [[ $status -ne 0 ]]; then
+if [[ ${IGNORE_IF_SRC_MISSING} == "true" ]]; then
```
sethyates: Why did you drop the `$?` approach?  It was more correct than do this approach.

sethyates: If you're wanting to not exit on error then you should change the `set -e` option

scripts/activities/gsutil-download.sh
arijitdey: This works fine. I tested it.

sethyates:
LGTM

PR #363, opened 2016-04-16
(Fixes #362) Don't emit empty arrays for EmrConfiguration

### LINKED ISSUE #362: EmrConfiguration emits empty arrays
When creating an EmrConfiguration and not supplying either sub-configurations or properties (or both), an empty array is emitted, which Data Pipeline does not allow. Here's the simplest (I think) pipeline that will create empty arrays:
```
import com.krux.hyperion.activity.SparkTaskActivity
import com.krux.hyperion.resource.{EmrConfiguration, SparkCluster}
import com.krux.hyperion.{DataPipelineDef, HyperionCli, HyperionContext, Schedule}
import com.typesafe.config.{ConfigFactory, ConfigValueFactory}

object EmptyArrayPipeline extends DataPipelineDef with HyperionCli {
  val config = ConfigFactory.load("reference")
    .withValue("hyperion.log.uri", ConfigValueFactory.fromAnyRef(""))
    .withValue("hyperion.script.uri", ConfigValueFactory.fromAnyRef(""))

  override implicit val hc: HyperionContext = new HyperionContext(config)

  val schedule = Schedule.onceAtActivation

  val emrConfig = EmrConfiguration()

  val cluster = SparkCluster().withConfiguration(emrConfig)

  override def workflow = SparkTaskActivity(s3 / "foo", "bar")(cluster)
}
```

This issue appears to be due to lines 30-31 in `EmrConfiguration`, where `Option` is explicitly used, preventing the logic in seq2Option from being applied.

Pull request coming shortly.

bakerlee:


realstraw:

PR #366, opened 2016-04-19
(Fixes #365) Removing bootstrap action when using emr label

### LINKED ISSUE #365: Change standard bootstrap action to use aws cli instead of hadoop
When we use release label 4.4.0 to create an EMR cluster, the standard bootstrap action needs to be configured using configurations.

arijitdey:
Native applications are no longer configured by bootstrap actions but by configurations in EMR release label 4.X

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/resource/EmrCluster.scala
+++ b/core/src/main/scala/com/krux/hyperion/resource/EmrCluster.scala
@@ -112,7 +112,12 @@ trait EmrCluster extends ResourceObject {
 
   def releaseLabel = emrClusterFields.releaseLabel
   def withReleaseLabel(label: HString): Self = updateEmrClusterFields(
-    emrClusterFields.copy(releaseLabel = Option(label), amiVersion = None)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/krux/hyperion/resource/EmrCluster.scala
```diff
@@ -112,7 +112,12 @@ trait EmrCluster extends ResourceObject {
 
   def releaseLabel = emrClusterFields.releaseLabel
   def withReleaseLabel(label: HString): Self = updateEmrClusterFields(
-    emrClusterFields.copy(releaseLabel = Option(label), amiVersion = None)
+    emrClusterFields.copy(
+      releaseLabel = Option(label),
+      amiVersion = None,
+      standardBootstrapAction = Seq.empty,
```
realstraw: You should channel `standardBootstrapAction` to the corresponding configuration rather than overwriting it to empty. The `standardBootstrapAction` made sure the clusters has a consistent setup, this implementation will break it.

PR #368, opened 2016-04-19
(Fixes #365) Change standard bootstrap action to use aws cli instead of hadoop

### LINKED ISSUE #365: Change standard bootstrap action to use aws cli instead of hadoop

arijitdey:
Change standard bootstrap action to use aws cli instead of hadoop


realstraw:
👍

PR #371, opened 2016-04-20
(Fixes #370) The standard bootstrap script now fails on older version of AMI (< 3.x)

### LINKED ISSUE #370: The standard bootstrap script now fails on older version of AMI (< 3.x)
Should check availability of command `aws`, if unavailable, fall back to use `hadoop`

antonwinter:
Adding support for ami versions < 4.x and EMR release label bootstrap actions.


realstraw:

PR #392, opened 2016-05-04
(Fixes #369) Making the SparkExample and SparkCluster compatible with…

### LINKED ISSUE #369: Issues using EMR4.3.0
I have a bunch of AWS DataPipelines using EMR4.3.0 with Spark 1.6.0 and naturally I want to programatically define them rather than manually writing up JSON definition files.

I just recently came across the Hyperion project which seems like something I could use. Just by looking at the Spark example and a bit of digging around though, it seems that the library is just supporting older style of data pipelines definitions. 

For example the apply methods of both SparkTaskActivity and SparkStep are instantiating their private companion classes with hard coded values for:
scriptRunner = "s3://elasticmapreduce/libs/script-runner/script-runner.jar",
jobRunner = s"${hc.scriptUri}run-spark-step.sh",

These are pre EMR 4.x.x ways of running spark. In EMR4.x.x it's recommended to use command-runner.jar and spark-submit e.g.:
```
{
 "name":"MyActivity",
 "step":[
  "command-runner.jar,",
  "spark-submit,",
  "--deploy-mode,cluster,",
  ...
 ]
 ...
}
```

I also noticed that in SparkCluster the way to specify Spark is via a standard bootstrap action s3://support.elasticmapreduce/spark/install-spark. The EMR 4.x.x way is to specify "applications":["spark"] in the EMRCluster object. 

There are other differences as well.

Some questions:
- Do you guys have plans to upgrade the DSL to support new EMR versions? When?
- Currently the configs I mentioned above are hard coded behind private case classes which cannot be directly instantiated by code using the library. Am I missing something. Is there another way to extend the functionality to allow generation of definitions for EMR4.x.x? Or do I need to modify the Hyperion code to do that?
- If modifying the Hyperion code is the only way forward, would you guys be willing to accept contributions in this direction?

asicoe:
… EMR4.3.

    Changes to be committed:

    modified:   core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
    Changes are minimal and they involve just overriding the standardBootstrapAction to an empty sequence and adding "spark" to the applications sequence. Both these changes are necessary because the new way of running Spark on EMR4.0+ does not require any bootstrap actions.

    modified:   contrib/activity/definition/src/test/resources/example.conf
    modified:   core/src/test/resources/example.conf
    modified:   examples/src/main/resources/example.conf
    Removed uneccesary entries like spark.version and added release label for emr4.3

    modified:   examples/src/test/scala/com/krux/hyperion/examples/ExampleMapReduceSpec.scala
    modified:   examples/src/test/scala/com/krux/hyperion/examples/ExampleS3DistCpWorkflowSpec.scala
    modified:   examples/src/test/scala/com/krux/hyperion/examples/ExampleSparkSpec.scala
    Fixed the spec to reflect above changes.

### DIFF
--- a/contrib/activity/definition/src/test/resources/example.conf
+++ b/contrib/activity/definition/src/test/resources/example.conf
@@ -16,8 +16,6 @@ hyperion {
             instance.type = "m3.xlarge"
             ami.version = "3.3"
             terminate = "8 hours"
-            spark.version = "1.1.1.e"
-            env.uri = "s3://bucket/org_env.sh"
         }
     }
 }
--- a/core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
+++ b/core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
@@ -12,8 +12,7 @@ import com.krux.hyperion.HyperionContext
 case class SparkCluster private (
   baseFields: BaseFields,
   resourceFields: ResourceFields,
-  emrClusterFields: EmrClusterFields,
-  sparkVersion: HString
+  emrClusterFields: EmrClusterFields
 ) extends EmrCluster {
 
   type Self = SparkCluster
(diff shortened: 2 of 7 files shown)


realstraw:
can we make this backwards compatible? only this when `cluster.releaseLabel.isDefined`?

sethyates:
Replaced this PR with #427

PR #395, opened 2016-05-13
(fixes #394) S3DistCp activity now accepts both s3 and hdfs URI

### LINKED ISSUE #394: S3DistCpActivity requires the destination to be S3, it should also accept HDFS

realstraw:


arijitdey:
👍

jnewman:
LGTM. Should there be an example w/ `s3` as the target?

realstraw:
@jnewman I think it's fine without s3 as target, we have s3 as source in the example anyway. Should be pretty self explanatory

PR #400, opened 2016-05-20
Issue398: No support for securityGroupIds in Ec2Resource

gtitievsky:


core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
```diff
@@ -69,7 +69,7 @@ object Ec2Resource {
     imageId = Option(hc.ec2ImageId: HString),
     runAsUser = None,
     associatePublicIpAddress = HBoolean.False,
-    securityGroups = Seq(hc.ec2SecurityGroup),
+    securityGroups = Seq.empty,
```
realstraw: This will break the default behavior, where people have the config setup expect the security group to be automatically put in. We would like to limit breaking changes in 3.x.x.

Can you make `hc.ec2SecurityGroup` an `Option` value, and `securityGroups` are set if `hc.ec2SecurityGroup` is defined. You can also do that with `securityGroupIds` as well, so it serves as default value and you do not need to have it everything you define your pipelines.

core/src/main/scala/com/krux/hyperion/HyperionContext.scala
```diff
@@ -35,6 +35,7 @@ class HyperionContext(config: Config) {
   lazy val ec2ResourceRole = Try(config.getString("hyperion.aws.ec2.resource.role")).toOption.getOrElse(resourceRole)
 
   lazy val ec2SecurityGroup = config.getString("hyperion.aws.ec2.securitygroup")
+  lazy val ec2SecurityGroupId = config.getString("hyperion.aws.ec2.securitygroupid")
```
realstraw: I don't see this get used anywhere, but as I suggested in the comment below. You might want to make this and `ec2SecurityGroup` an `Option`, and have them filled in as default (if defined) to the ec2 security group or group id.

gtitievsky:
Please see if you can resolve the conflicts?

gtitievsky:
@realstraw: could you please check the latest checkins?

core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
```diff
@@ -69,8 +69,8 @@ object Ec2Resource {
     imageId = Option(hc.ec2ImageId: HString),
     runAsUser = None,
     associatePublicIpAddress = HBoolean.False,
-    securityGroups = Seq(hc.ec2SecurityGroup),
-    securityGroupIds = Seq.empty,
+    securityGroups = if (hc.ec2SecurityGroup.isEmpty) Seq.empty else Seq(hc.ec2SecurityGroup.get),
```
realstraw: you can simply use `hc.ec2SecurityGroup.toSeq`

core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
```diff
@@ -69,8 +69,8 @@ object Ec2Resource {
...
+    securityGroupIds = if (hc.ec2SecurityGroupId.isEmpty) Seq.empty else Seq(hc.ec2SecurityGroupId.get),
```
realstraw: you can simply use `hc.ec2SecurityGroupId.toSeq`

gtitievsky:
@realstraw : better?

realstraw:

PR #402, opened 2016-05-22
(fixes #401) Handle .compare(_) on parameters without default values

### LINKED ISSUE #401: Parameter to HType conversion should not be Evaluatable when default not specified
A Parameter[+OptionalOrdered] without a default value that is then compared will throw an exception. Here's an example:
```
val instanceCount = Parameter[Int]("InstanceCount")
SparkCluster().withCoreInstanceCount(instanceCount)
```

bakerlee:


realstraw:
Does the following still hold with this implementation?

```scala
(Parameter[Int]("Param").withValue(1) > 0) === Some(true)
(Parameter[Int]("Param").withValue(1) < 0) === Some(false)
(Parameter("Param").encrypted(1) > 0) === Some(true)
(Parameter("Param").encrypted(1) < 0) === Some(false)
```

bakerlee:
Yes, it still holds. I had to change the last two slightly, though, to this:
```
(Parameter[Int]("Param").encrypted.withValue(1) > 0) === Some(true)
(Parameter[Int]("Param").encrypted.withValue(1) < 0) === Some(false)
```
I've added these to the unit test as well.

realstraw:
👍 Thanks, @bakerlee. can you resolve the conflicts? I'll merge after.

PR #412, opened 2016-06-21
(fixes #404) PythonActivity script now uses the correct virtualenv path

### LINKED ISSUE #404: PythonActivity script does not set up the virtualenv correctly
Need to use `.ENV/bin/activate` instead of `bin/activate`

realstraw:


jpurnell01:
👍  thanks!

PR #413, opened 2016-06-21
(fixes #403) update the default ec2 ami to the latest

### LINKED ISSUE #403: Update the default to use the latest amazon AMIs for ec2 resources
https://aws.amazon.com/amazon-linux-ami/

and also update the default ec2 instance type to the more modern t2.small instead of the previous generation m1.small

realstraw:


jpurnell01:
LGTM

PR #414, opened 2016-06-22
(fixes #322) making note that name field does not accept expressions

### LINKED ISSUE #322: Verify if the `name` field can use expression or not and change accordingly
Not sure if `name` can include expressions. If it can need to change the type of `name` type from `Option[String]` to `Option[HString]`

realstraw:


jpurnell01:
:+1:

PR #419, opened 2016-06-29
(Fixes #418) Use configurable exponential backoff + jitter for retry

### LINKED ISSUE #418: Max retry should be configurable and use exponential backoff and jitter instead of fixed interval

realstraw:


jpurnell01:
👍

PR #422, opened 2016-06-30
(Fixes #356) escape `,` in arguments of EMR and Spark step arguments

### LINKED ISSUE #356: Need to escape `,` in arguments of Emr and Spark steps

realstraw:


arijitdey:
👍

PR #425, opened 2016-07-06
Fixes #380

### LINKED ISSUE #380: Make SparkCluster and SparkStep compatible to EMR release label 4.x
Currently spark cluster in using bootstrap actions to install spark, which worked fine in AMI 3.x, but not in release label 4.x. It needs to be updated to be able to work with release label 4.4 through configuration and change SparkStep accordingly.

hoangelos:


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+
```
sethyates: Please add comments on your change to the changelog.

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
```diff
@@ -13,19 +14,24 @@ import com.krux.hyperion.resource.{ Resource, SparkCluster }
...
   activityFields: ActivityFields[SparkCluster],
+  jobRunner: HString,
+  scriptRunner: HString,
   steps: Seq[SparkStep],
   inputs: Seq[S3DataNode],
   outputs: Seq[S3DataNode],
   preStepCommands: Seq[HString],
   postStepCommands: Seq[HString]
-) extends EmrActivity[SparkCluster] {
+)(implicit hc: HyperionContext) extends EmrActivity[SparkCluster] {
```
sethyates: We've tried to avoid passing the implicit to the case class constructor.  The `apply` should be sufficient for this?

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
```diff
@@ -59,9 +65,27 @@ case class SparkActivity private (
 
 object SparkActivity extends RunnableObject {
 
-  def apply(runsOn: Resource[SparkCluster]): SparkActivity = new SparkActivity(
+  def jobRunner(runsOn: Resource[SparkCluster])(implicit hc: HyperionContext): HString = {
+    if(runsOn.asManagedResource.exists(_.isReleaseLabel4xx)) {
```
sethyates: Space between `if` and `(`.

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
```diff
@@ -59,9 +65,27 @@ case class SparkActivity private (
...
+      "command-runner.jar"
+    } else {
+      "s3://elasticmapreduce/libs/script-runner/script-runner.jar"
+    }
+  }
+
+  def apply(runsOn: Resource[SparkCluster])(implicit hc: HyperionContext): SparkActivity = new SparkActivity(
     baseFields = BaseFields(PipelineObjectId(SparkActivity.getClass)),
     activityFields = ActivityFields(runsOn),
+    jobRunner = jobRunner(runsOn)(hc),
```
sethyates: Explicitly passing `hc` shouldn't be required?

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
```diff
@@ -11,12 +11,15 @@ case class SparkStep private (
   jarUri: HS3Uri,
   mainClass: Option[MainClass],
   args: Seq[HString],
-  scriptRunner: HString,
-  jobRunner: HString,
+  var scriptRunner: HString,
```
sethyates: Why are these marked as `var` now?

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
```diff
@@ -51,8 +54,8 @@ object SparkStep {
     jarUri = jarUri,
     mainClass = None,
     args = Seq.empty,
-    scriptRunner = "s3://elasticmapreduce/libs/script-runner/script-runner.jar",
-    jobRunner = s"${hc.scriptUri}run-spark-step.sh",
+    scriptRunner = "",
```
sethyates: More preferrably these would be marked as `Option[HString]` and defaulted to `None` rather than empty string.

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
```diff
@@ -11,12 +11,15 @@ case class SparkStep private (
...
-  scriptRunner: HString,
-  jobRunner: HString,
+  var scriptRunner: HString,
+  var jobRunner: HString,
   sparkOptions: Seq[HString],
   sparkConfig: Map[HString, HString]
 ) {
 
+
+  def withJobRunner(runner: HString) = copy(jobRunner = runner)
```
sethyates: `copy(jobRunner = Option(runner))` assuming you changed the property to `jobRunner: Option[HString]` as per below.

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
```diff
@@ -29,7 +27,7 @@ case class SparkTaskActivity private (
   outputs: Seq[S3DataNode],
   sparkOptions: Seq[HString],
   sparkConfig: Map[HString, HString]
-) extends EmrTaskActivity[SparkCluster] {
+)(implicit hc: HyperionContext) extends EmrTaskActivity[SparkCluster] {
```
sethyates: See comment about regarding implicit on the case class constructor.

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
```diff
@@ -55,6 +53,22 @@ case class SparkTaskActivity private (
   def withFiles(files: HString*) = withSparkOption(files.flatMap(file => Seq("--files": HString, file)): _*)
   def withMaster(master: HString) = withSparkOption("--master", master)
 
+  def jobRunner: HString = {
```
sethyates: See comments above on `SparkActivity`.  Would be nice to DRY this up between these two if possible.

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -20,15 +20,32 @@ case class SparkCluster private (
...
   def updateResourceFields(fields: ResourceFields) = copy(resourceFields = fields)
   def updateEmrClusterFields(fields: EmrClusterFields) = copy(emrClusterFields = fields)
 
   def withSparkVersion(sparkVersion: HString) = copy(sparkVersion = sparkVersion)
 
+  override def withReleaseLabel(label: HString): Self = {
+    super.withReleaseLabel(label)
+    if (label.toString.startsWith("emr-4.")) {
+      this.withApplications("Spark": HString)
+      this.defaultBootstrapActions = Seq.empty[HString]
```
sethyates: Rather than marking all of these as `var` and making this no longer an immutable object, why don't you simply `this.copy(defaultApplications = x, defaultBootstrapActions = y, etc)`?

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -20,15 +20,32 @@ case class SparkCluster private (
 
   val logger = LoggerFactory.getLogger(SparkCluster.getClass)
 
+  var isReleaseLabel4xx = false
```
sethyates: And what happens when there is a release label `5.x`?

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
```diff
@@ -25,7 +28,15 @@ case class SparkActivity private (
   def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
   def updateActivityFields(fields: ActivityFields[SparkCluster]) = copy(activityFields = fields)
 
-  def withSteps(step: SparkStep*) = copy(steps = steps ++ step)
+  def withSteps(step: SparkStep*) = {
+    val new_steps = step.map(item =>
```
sethyates: Please use camel case so `newSteps` not `new_steps`.

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
```diff
@@ -25,7 +28,15 @@ case class SparkActivity private (
...
+      item.copy(
+        scriptRunner = Some(scriptRunner),
```
sethyates: We prefer `Option` as a constructor and `Some` used only for matching.

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
```diff
@@ -34,7 +37,7 @@ case class SparkStep private (
   def withMaster(master: HString) = withSparkOption("--master", master)
 
   def serialize: String = Seq(
-    Seq(scriptRunner, jobRunner),
+    Seq(scriptRunner.getOrElse(""), jobRunner.getOrElse("")),
```
sethyates: Could you not simply have:
```
Seq(
  scriptRunner,
  jobRunner,
  sparkOptions,
  ...
)
```

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
```diff
@@ -47,12 +50,18 @@ case class SparkStep private (
...
 object SparkStep {
 
-  def apply(jarUri: HS3Uri)(implicit hc: HyperionContext): SparkStep = SparkStep(
+  def apply(jarUri: HS3Uri)(implicit hc: HyperionContext): SparkStep = apply(
+    jarUri = jarUri,
+    scriptRunner = None,
+    jobRunner = None
+  )
+
+  def apply(jarUri: HS3Uri, jobRunner: Option[HString], scriptRunner: Option[HString])(implicit hc: HyperionContext): SparkStep = SparkStep(
```
sethyates: Any reason not to have `jobRunner: Option[HString] = None, scriptRunner: Option[HString] = None)` rather than this and a separate `apply` without these parameters?

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
```diff
@@ -47,12 +50,18 @@ case class SparkStep private (
...
     jarUri = jarUri,
     mainClass = None,
     args = Seq.empty,
-    scriptRunner = "s3://elasticmapreduce/libs/script-runner/script-runner.jar",
-    jobRunner = s"${hc.scriptUri}run-spark-step.sh",
+    scriptRunner = None,
```
sethyates: Why aren't you passing in the parameters here and the next line?

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
```diff
@@ -19,8 +19,8 @@ case class SparkTaskActivity private (
   baseFields: BaseFields,
   activityFields: ActivityFields[SparkCluster],
   emrTaskActivityFields: EmrTaskActivityFields,
-  scriptRunner: HString,
   jobRunner: HString,
+  scriptRunner: HString,
```
sethyates: Why was this moved down?

core/src/main/scala/com/krux/hyperion/common/SparkCommandRunner.scala
```diff
@@ -0,0 +1,24 @@
+package com.krux.hyperion.common
+
+import com.krux.hyperion.HyperionContext
+import com.krux.hyperion.adt.HString
+import com.krux.hyperion.resource.{SparkCluster, Resource}
+
+trait SparkCommandRunner {
```
sethyates: Thanks!

core/src/main/scala/com/krux/hyperion/common/SparkCommandRunner.scala
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  def scriptRunner(runsOn: Resource[SparkCluster]): HString = {
+    if (runsOn.asManagedResource.exists(_.isReleaseLabel4xx)) {
+      "command-runner.jar"
+    } else {
+      "s3://elasticmapreduce/libs/script-runner/script-runner.jar"
+    }
+  }
+}
```
sethyates: Newline please.

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -26,9 +30,25 @@ case class SparkCluster private (
...
   def withSparkVersion(sparkVersion: HString) = copy(sparkVersion = sparkVersion)
 
+  override def withReleaseLabel(label: HString): Self = {
+    super.withReleaseLabel(label)
+    if ((label.toString.length >= 6) && (label.toString.substring(4, 5).toInt >= 4)) {
+      //Assuming emr-x.y.z, and return x as a number.
+      this.copy(
+        defaultApplications = Seq("Spark": HString),
+        defaultBootstrapActions = Seq.empty[HString],
+        isReleaseLabel4xx = true
```
sethyates: I still contend this is a bad name for this variable.  Doesn't the mere presence of having a releaseLabel trigger the different behaviour?  So this would be redundant and you could just use `releaseLabel.nonEmpty`, right?  I also don't think you need to do the string checking above as you wouldn't be getting this called unless the callee were using a release label.

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
hoangelos: The reason I added a separate was to allow people to override the default. Although it's not possible so it does make sense. Or else I could change it to ignore setting it unless it is None.

sethyates: Yes but you can do the same thing with a single `apply` like so:

```
def apply(jarUri: HS3Uri, jobRunner: Option[HString] = None, scriptRunner: Option[HString] = None)(implicit hc: HyperionContext): SparkStep = ...
```

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
hoangelos: Technically it isn't just the lack of the exist an e but at least 4.x. Or greater right?

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
hoangelos: Of course. But what do you think about SparkActivity not updating it if SoarkStep already is set?  That way users can overwrite with their own?

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
sethyates: Release labels only supported >= 4.x.  My reading of this is pre-release labels (aka pre 4.x), we did one thing, afterwards, we do it a different way.

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
sethyates: OK but this `apply` is about creating new `SparkStep`s.  The not-updating would be handled in `SparkActivity` itself.  I'll post a comment there on how it would appear.  I'm actually unsure if this is a real use-case though...

hoangelos: It is for us. ;)

core/src/main/scala/com/krux/hyperion/activity/SparkActivity.scala
sethyates: If you wanted to keep what was specified in the SparkStep and only apply these as a "default", you could use: `item.copy(scriptRunner = item.scriptRunner.orElse(Option(scriptRunner)), ...)`.

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
sethyates: Cool! :-)

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
hoangelos: If I do that it outputs the Option instead of the value, which would affect the JSON wouldn't it?

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
hoangelos: I can convert to using that.  I assumed this was something inherent in emr-4.x.x, but you are saying it's more backwards compatibility with hyperion library.

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -13,7 +13,11 @@ case class SparkCluster private (
   baseFields: BaseFields,
   resourceFields: ResourceFields,
   emrClusterFields: EmrClusterFields,
-  sparkVersion: HString
+  sparkVersion: HString,
+  defaultApplications: Seq[HString],
```
sethyates: defaultApplications, defaultBootstrapActions and hasReleaseLabel are not required.

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -26,9 +30,24 @@ case class SparkCluster private (
 
   def withSparkVersion(sparkVersion: HString) = copy(sparkVersion = sparkVersion)
 
+  override def withReleaseLabel(label: HString): Self = {
```
sethyates: Remove this.

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -26,9 +30,24 @@ case class SparkCluster private (
...
+        defaultApplications = Seq("Spark": HString),
+        defaultBootstrapActions = Seq.empty[HString],
+          hasReleaseLabel = true
+      )
+    } else {
+      this
+    }
+  }
+
+  override def applications =
```
sethyates:
```
override def applications = if (releaseLabel.nonEmpty)
  ("Spark": HString) +: super.applications
else
  super.applications
```

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -26,9 +30,24 @@ case class SparkCluster private (
...
+    this.defaultApplications ++ super.applications
+
   override def standardBootstrapAction = 
-    (s"s3://support.elasticmapreduce/spark/install-spark,-v,${sparkVersion},-x": HString) +:
-    super.standardBootstrapAction
+    this.defaultBootstrapActions ++ super.standardBootstrapAction
```
sethyates:
```
override def standardBootstrapAction = if (releaseLabel.nonEmpty)
  super.standardBootstrapAction
else
  (s"s3://support.elasticmapreduce/spark/install-spark,-v,${sparkVersion},-x": HString) +: super.standardBootstrapAction
```

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -38,7 +57,10 @@ object SparkCluster {
     baseFields = BaseFields(PipelineObjectId(SparkCluster.getClass)),
     resourceFields = EmrCluster.defaultResourceFields(hc),
     emrClusterFields = EmrCluster.defaultEmrClusterFields(hc),
-    sparkVersion = hc.emrSparkVersion.get
+    sparkVersion = hc.emrSparkVersion.get,
+    defaultApplications = Seq.empty[HString],
```
sethyates: defaultApplications, defaultBootstrapActions and hasReleaseLabel are not required.

core/src/main/scala/com/krux/hyperion/common/SparkCommandRunner.scala
```diff
@@ -0,0 +1,24 @@
+package com.krux.hyperion.common
+
+import com.krux.hyperion.HyperionContext
+import com.krux.hyperion.adt.HString
+import com.krux.hyperion.resource.{SparkCluster, Resource}
+
+trait SparkCommandRunner {
+
+  def jobRunner(runsOn: Resource[SparkCluster])(implicit hc: HyperionContext): HString = {
+    if (runsOn.asManagedResource.exists(_.hasReleaseLabel)) {
```
sethyates:
```
if (runsOn.asManagedResource.exists(_.releaseLabel.nonEmpty)) {
```

core/src/main/scala/com/krux/hyperion/common/SparkCommandRunner.scala
```diff
@@ -0,0 +1,24 @@
...
+      "spark-submit"
+    } else {
+      s"${hc.scriptUri}run-spark-step.sh"
+    }
+  }
+
+  def scriptRunner(runsOn: Resource[SparkCluster]): HString = {
+    if (runsOn.asManagedResource.exists(_.hasReleaseLabel)) {
```
sethyates:
```
if (runsOn.asManagedResource.exists(_.releaseLabel.nonEmpty)) {
```

core/src/main/scala/com/krux/hyperion/activity/SparkStep.scala
sethyates: Got it.

sethyates:
Hi @hoangelos, I've looked into this more and I think the initial approach won't cover all cases (such as for some reason setting `withAmiVersion` after using `withReleaseLabel`).  I've made some recommendations above that should actually simplify this and handle all these cases.  Let me know what you think.

sethyates:
Replaced this PR with #427

Issue #375, opened 2016-04-21
Implement a PgpDecryptActivity


sethyates:
Using (probably) OpenPGP and BouncyCastle, implement a contrib activity to PGP decrypt a file.

sethyates:
Another option is using the pgp unix command assuming the Amazon AMI's have a reasonable version of it.

sethyates:
This should decrypt every non-hidden input file and produce a decrypted file in the output folder.  If an input file is not encrypted, it should be copied across untouched.  Optionally support ascii armored input and producing a _SUCCESS marker.

prjasper:
I've submitted an initial implementation in this PR #426
Feedback welcome.
It doesn't add a _SUCCESS marker or filter out hidden files yet, but would be relatively easy to add if these are important. It doesn't copy across unencrypted files.
It's implemented using the GNU gpg command.

prjasper:
Now creates a _SUCCESS marker on successful decryption.
Also filters out hidden files and directories, and empty/marker files like _SUCCESS.
(Still doesn't copy across unencrypted files.)

PR #426, opened 2016-07-07
(Fixes #374 & #375) Add activities to encrypt and decrypt files using Gnu PGP encryption.

### LINKED ISSUE #374: Implement a PgpEncryptActivity
Using (probably) OpenPGP and BouncyCastle, implement a contrib activity to PGP encrypt a file.

prjasper:
This PR adds Hyperion activities for encrypting and decrypting files using GNU PGP encryption. I'm expecting comments on this first PR.

### DIFF
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
@@ -34,7 +34,7 @@ case class PgpDecryptActivity private(
   def markOnSuccess = copy(markSuccessfulJobs = true)
 
   override def scriptArguments = Seq(
-    if (markSuccessfulJobs) Option("--markSuccessfulJobs") else None,
+    if (markSuccessfulJobs) Option("--mark-successful-jobs") else None,
     Option(key.serialize)
   ).flatten
 }
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
@@ -34,8 +34,8 @@ case class PgpDecryptActivity private(
   def markOnSuccess = copy(markSuccessfulJobs = true)
 
   override def scriptArguments = Seq(
-    if (markSuccessfulJobs) Some("--markSuccessfulJobs") else None,
-    Some(key.serialize)
+    if (markSuccessfulJobs) Option("--markSuccessfulJobs") else None,
+    Option(key.serialize)
   ).flatten
 }
 
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
@@ -15,11 +15,12 @@ import com.krux.hyperion.resource.{Ec2Resource, Resource}
   * @param shellCommandActivityFields the shell command setup fields
   * @param key the file containing the private decryption key
   */
-case class PgpDecryptActivity private(baseFields: BaseFields,
-                                      activityFields: ActivityFields[Ec2Resource],
-                                      shellCommandActivityFields: ShellCommandActivityFields,
-                                      key: HS3Uri)
-  extends PgpActivity with WithS3Input with WithS3Output {
+case class PgpDecryptActivity private(
+  baseFields: BaseFields,
+  activityFields: ActivityFields[Ec2Resource],
+  shellCommandActivityFields: ShellCommandActivityFields,
+  key: HS3Uri
+) extends PgpActivity with WithS3Input with WithS3Output {
   type Self = PgpDecryptActivity
 
   def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
@@ -33,9 +34,8 @@ case class PgpDecryptActivity private(baseFields: BaseFields,
 
 object PgpDecryptActivity
   extends RunnableObject {
-  def apply(key: HS3Uri)
-           (runsOn: Resource[Ec2Resource])
-           (implicit hc: HyperionContext): PgpActivity = PgpDecryptActivity(
+  def apply(key: HS3Uri)(runsOn: Resource[Ec2Resource])(implicit hc: HyperionContext): PgpActivity = PgpDecryptActivity(
     baseFields = BaseFields(PipelineObjectId(PgpDecryptActivity.getClass)), activityFields = ActivityFields(runsOn),
-    shellCommandActivityFields = ShellCommandActivityFields(PgpActivity.decryptScript), key = key)
+    shellCommandActivityFields = ShellCommandActivityFields(PgpActivity.decryptScript), key = key
+  )
 }
--- a/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
+++ b/contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
@@ -1,7 +1,7 @@
 package com.krux.hyperion.activity
 
 import com.krux.hyperion.HyperionContext
-import com.krux.hyperion.adt.{HS3Uri, HString}
+import com.krux.hyperion.adt.{HBoolean, HS3Uri, HString}
 import com.krux.hyperion.common.{BaseFields, PipelineObjectId}
 import com.krux.hyperion.expression.RunnableObject
 import com.krux.hyperion.resource.{Ec2Resource, Resource}
@@ -14,12 +14,14 @@ import com.krux.hyperion.resource.{Ec2Resource, Resource}
   * @param activityFields the activity setup fields
   * @param shellCommandActivityFields the shell command setup fields
   * @param key the file containing the private decryption key
+  * @param markSuccessfulJobs add a _SUCCESS file to the output location on success
   */
 case class PgpDecryptActivity private(
   baseFields: BaseFields,
   activityFields: ActivityFields[Ec2Resource],
   shellCommandActivityFields: ShellCommandActivityFields,
-  key: HS3Uri
+  key: HS3Uri,
+  markSuccessfulJobs: HBoolean
 ) extends PgpActivity with WithS3Input with WithS3Output {
   type Self = PgpDecryptActivity
 
@@ -29,13 +31,19 @@ case class PgpDecryptActivity private(
 
   def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
 
-  override def scriptArguments = Seq(key.serialize: HString)
+  def markOnSuccess = copy(markSuccessfulJobs = true)
+
+  override def scriptArguments = Seq(markSuccessfulJobs.serialize, key.serialize: HString)
 }
 
 object PgpDecryptActivity
   extends RunnableObject {
-  def apply(key: HS3Uri)(runsOn: Resource[Ec2Resource])(implicit hc: HyperionContext): PgpActivity = PgpDecryptActivity(
-    baseFields = BaseFields(PipelineObjectId(PgpDecryptActivity.getClass)), activityFields = ActivityFields(runsOn),
-    shellCommandActivityFields = ShellCommandActivityFields(PgpActivity.decryptScript), key = key
-  )
+  def apply(key: HS3Uri)(runsOn: Resource[Ec2Resource])(implicit hc: HyperionContext): PgpActivity =
+    new PgpDecryptActivity(
+      baseFields = BaseFields(PipelineObjectId(PgpDecryptActivity.getClass)),
+      activityFields = ActivityFields(runsOn),
+      shellCommandActivityFields = ShellCommandActivityFields(PgpActivity.decryptScript),
+      key = key,
+      markSuccessfulJobs = HBoolean.False
+    )
 }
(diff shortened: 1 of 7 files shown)


contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
```diff
@@ -0,0 +1,41 @@
...
+  * The server-side PGP decryption activity decrypts files from the input location to the output location using the
+  * private decryption key.
+  *
+  * @param baseFields the pipeline base fields
+  * @param activityFields the activity setup fields
+  * @param shellCommandActivityFields the shell command setup fields
+  * @param key the file containing the private decryption key
+  */
+case class PgpDecryptActivity private(baseFields: BaseFields,
+                                      activityFields: ActivityFields[Ec2Resource],
```
sethyates: Align the fields to one "tabstop" rather than aligning to the open parens.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  override def scriptArguments = Seq(key.serialize: HString)
+}
+
+object PgpDecryptActivity
+  extends RunnableObject {
+  def apply(key: HS3Uri)
+           (runsOn: Resource[Ec2Resource])
```
sethyates: Alignment.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpEncryptActivity.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  override def scriptArguments = Seq(key.serialize: HString)
+}
+
+object PgpEncryptActivity
+  extends RunnableObject {
+  def apply(key: HS3Uri)
+           (runsOn: Resource[Ec2Resource])
```
sethyates: Alignment.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpEncryptActivity.scala
```diff
@@ -0,0 +1,41 @@
...
+  * The server-side PGP encryption activity encrypts files from the input location to the output location using the
+  * public encryption key.
+  *
+  * @param baseFields the pipeline base fields
+  * @param activityFields the activity setup fields
+  * @param shellCommandActivityFields the shell command setup fields
+  * @param key the file containing the public encryption key
+  */
+case class PgpEncryptActivity private(baseFields: BaseFields,
+                                      activityFields: ActivityFields[Ec2Resource],
```
sethyates: Alignment.

scripts/activities/gpg-decrypt.sh
```diff
@@ -0,0 +1,93 @@
...
+  if [[ ${n} -eq 0 ]]; then
+    trap on_exit EXIT
+  fi
+}
+
+# get a file, whether local or on S3
+# $1 the path to the unencrypted file
+# $2 the working directory
+function get_file() {
+  LOCAL="$2"/"$(basename "$1")"
```
sethyates: So how does this work when `$1` is `s3://foo/bar`?

scripts/activities/gpg-decrypt.sh
```diff
@@ -0,0 +1,93 @@
...
+IFS=$'\n\t'
+
+# create a temporary working directory
+if [ -x /usr/local/bin/gmktemp ]; then
+  # for testing on MacOS
+  MKTEMP=/usr/local/bin/gmktemp
+else
+  MKTEMP=mktemp
+fi
+WORKING_DIR=$($MKTEMP -d)
```
sethyates: Just move `WORKING_DIR` up into the `if`/`else` above without the `MKTEMP` bit.

scripts/activities/gpg-decrypt.sh
sethyates: nm - didn't read that correctly.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
prjasper: Do you have a preferred set of IntelliJ Scala format settings? I'm finding it hard to get a combination that doesn't make everything really spread out.
(File > Other Settings > Default Settings... > Editor > Code Style > Scala > Manage > Export... would be nice, if you have a moment!)

scripts/activities/gpg-decrypt.sh
prjasper: Yeah LOCAL is just where it ends up.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
sethyates: Sent.

scripts/activities/gpg-encrypt.sh
```diff
@@ -0,0 +1,98 @@
+#! /usr/bin/env bash
```
sethyates: Same comments here as in the decrypt one.

scripts/activities/gpg-decrypt.sh
```diff
@@ -0,0 +1,92 @@
...
+if [ -x /usr/local/bin/gmktemp ]; then
+  # for testing on MacOS
+  WORKING_DIR="$(gmktemp -d)"
+else
+  WORKING_DIR="$(mktemp -d)"
+fi
+add_on_exit rm -rf \"${WORKING_DIR}\"
+
+# the file(s) to decrypt
+FILES=($(find "${INPUT1_STAGING_DIR}" -type f))
```
sethyates: Please use `-type f -not -empty` and I think this will be pretty close to meeting the requirement of skipping "hidden" files.

scripts/activities/gpg-decrypt.sh
```diff
@@ -0,0 +1,92 @@
+#! /usr/bin/env bash
+
+usage() {
+  NAME=$(basename $0)
+  cat <<EOF
+Usage: ${NAME} KEY
```
sethyates: I'd like to go ahead an account for creating the _SUCCESS marker file now.  This has proved very useful in other cases to have this as an option.

```
MARK_SUCCESSFUL_JOBS=0

while [[ $# > 0 ]]; do
  key=$1; shift

  case ${key} in
    --mark-successful-jobs)
      MARK_SUCCESSFUL_JOBS=1;
      ;;

    --key)
      KEY_FILE="$1"; shift
      ;;

    *)
      echo "Unrecognized option ${key}"
      exit 3
      ;;
  esac
done

...

if [ ${MARK_SUCCESSFUL_JOBS} -eq 1 ]; then
  touch ${OUTPUT2_STAGING_DIR}/_SUCCESS
fi
```

scripts/activities/gpg-decrypt.sh
prjasper: By "hidden" you mean files and directories beginning with '.'? I don't think ```-not -empty``` will avoid those.
Maybe ```-name ".*" -prune -o -type f```?

sethyates: Hence my "pretty close".  The "hidden" files I wanted to skip would be empty anyway (like _SUCCESS).

Suppose we could do `-type f -not -empty -not -name ".*" -not -name "_*"`

prjasper: Then maybe ```-name ".*" -prune -o -type f -not -empty -not -name "_*"``` which should also avoid files in hidden directories (like ```.foo/bar```).

sethyates: OK

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpEncryptActivity.scala
```diff
@@ -0,0 +1,48 @@
...
+
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def markOnSuccess = copy(markSuccessfulJobs = true)
+
+  override def scriptArguments = Seq(markSuccessfulJobs.serialize, key.serialize: HString)
```
sethyates: This is incorrect. It will generate 'true s3://bucket/key'. You want to add '--mark-successful-jobs' not the Boolean.

scripts/activities/gpg-decrypt.sh
```diff
@@ -0,0 +1,111 @@
...
+# process options
+MARK_SUCCESSFUL_JOBS=0
+while [[ $# > 1 ]]; do
+  case "$1" in
+    --mark-successful-jobs)
+      MARK_SUCCESSFUL_JOBS=1
+      shift
+      ;;
+    *)
+      break
```
sethyates: This will be where you pick up the mandatory key.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
```diff
@@ -0,0 +1,52 @@
...
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def markOnSuccess = copy(markSuccessfulJobs = true)
+
+  override def scriptArguments = Seq(
+    if (markSuccessfulJobs) Some("--markSuccessfulJobs") else None,
```
sethyates: We typically use `Option` as the constructor and `Some` only for pattern matching.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpEncryptActivity.scala
```diff
@@ -0,0 +1,51 @@
...
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def markOnSuccess = copy(markSuccessfulJobs = true)
+
+  override def scriptArguments = Seq(
+    if (markSuccessfulJobs) Some("--markSuccessfulJobs") else None,
```
sethyates: Same comments as above.

scripts/activities/gpg-decrypt.sh
prjasper: Line #17 drops out of the loop when there's one argument left, which gets processed after the loop. (I'm using the loop for processing options only -- imagine where there are multiple mandatory/positional arguments.) Line #23 drops out of the loop when an unrecognized option is encountered, but could give an error/usage message. Does that work for you?

sethyates: As you have it now, `gpg-decrypt.sh --mark-successful-jobs` would pass over this, pass by line 29 and fail on or after line 96 in possibly difficult-to-determine ways.  I'd change this to do all the argument parsing here so it's clearer what's happening and change the check on line 29 to ensure `$KEY_PATH` is set in the option parsing.

so:
```
-*)
  echo "Unrecognized option: $1"
  usage
  ;;

*)
  KEY_PATH=$1
  shift
  ;;
esac
done

if [ -z "${KEY_PATH}" ]; then
  echo "ERROR: wrong number of arguments"
  usage
fi
```

prjasper: The ```-*)``` case is good, but doing the mandatory arguments in the loop isn't easy to generalize to multiple arguments.) (But I'm happy that we're only debating minutiae, not something majorly wrong with the whole script.)

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpEncryptActivity.scala
```diff
@@ -0,0 +1,51 @@
...
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def markOnSuccess = copy(markSuccessfulJobs = true)
+
+  override def scriptArguments = Seq(
+    if (markSuccessfulJobs) Option("--markSuccessfulJobs") else None,
```
sethyates: `Option("--mark-successful-jobs")`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/PgpDecryptActivity.scala
```diff
@@ -0,0 +1,52 @@
...
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def markOnSuccess = copy(markSuccessfulJobs = true)
+
+  override def scriptArguments = Seq(
+    if (markSuccessfulJobs) Option("--markSuccessfulJobs") else None,
```
sethyates: `Option("--mark-successful-jobs")`

sethyates:
LGTM

sethyates:
:+1:

PR #427, opened 2016-07-09
(Fixes #369, Fixes #380, Closes #425, Closes #392) Make SparkCluster and SparkStep compatible to EMR release label 4.x

### LINKED ISSUE #369: Issues using EMR4.3.0

### LINKED ISSUE #380: Make SparkCluster and SparkStep compatible to EMR release label 4.x

sethyates:


core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
```diff
@@ -13,7 +13,7 @@ case class SparkCluster private (
   baseFields: BaseFields,
   resourceFields: ResourceFields,
   emrClusterFields: EmrClusterFields,
-  sparkVersion: HString
+  sparkVersion: Option[HString]
```
realstraw: If I understand correctly, this field is only used for non-release label clusters?

realstraw:
build.sbt is not updated with the new version

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 3.5.3 - 2016-07-09
```
realstraw: 3.6.0 maybe?

core/src/main/scala/com/krux/hyperion/resource/SparkCluster.scala
sethyates: Yes AFAICT

CHANGELOG.md
sethyates: Indeed

realstraw:
👍

PR #428, opened 2016-07-09
(Fixes #333) Allow explicitly converting an Activity to a WorkflowExpression

### LINKED ISSUE #333: Allow explicitly converting an Activity to a WorkflowExpression
Right now, you can declare an activity and then add a type declaration to trigger an implicit conversion (or call the `apply` of WorkflowExpression directly).  It would be nice to have a `.toWorkflowExpression` method on the base `Activity` class, so we could do something like the following without a type declaration/implicit:

```
val doSomething = if (someCondition) JarActivity(...).toWorkflowExpression else WorkflowNoActivityExpression
```

Right now, this would be accomplished using a type decl and implicit:

```
val doSomething: WorkflowExpression = if (someCondition) JarActivity(...) else WorkflowNoActivityExpression
```

sethyates:


realstraw:
👍

PR #429, opened 2016-07-10
(Fixes #343 Fixes #379 Fixes #393) Support ondemand schedules and customising DefaultObject

### LINKED ISSUE #343: Support the new `ondemand` schedule type
https://docs.aws.amazon.com/datapipeline/latest/DeveloperGuide/dp-concepts-schedules.html

### LINKED ISSUE #379: Cannot override DefaultObject
Currently, `DefaultObject` is a case class with a schedule and a hard-coded map.  This makes it impossible to override even though it is declared as `def defaultObject: DefaultObject` in `DataPipelineDef`.

### LINKED ISSUE #393: Ability to add custom properties to pipeline objects
This feature is similar to https://github.com/krux/hyperion/issues/379

Both of them are useful when creating large data pipelines with many parameters because they allow defining a number of parameters in the default object and then combining them in other parameters inside activity objects. This allows better structuring and also to overcome the max 50 top level parameter limitation of data pipeline.

E.g:
```
{
 "parameters" : [
  {"id" : "my_app1_name", "description" : "app1 name", "type" : "String", "default" : "latest"},
  {"id" : "my_app2_name", "description" : "app2 name", "type" : "String", "default" : "latest"},
  {"id" : "my_app3_name", "description" : "app3 name", "type" : "String", "default" : "latest"}
],
"objects" : [
 {"id" : "Default", "my_app_jar_bucket" : "bucket"},
 {"id" : "Activity1", "type" : "EmrActivity", "my_app_jar" : "s3://#{my_app_jar_bucket}/#{my_app1_name}.jar", "step" : [ "command-runner.jar", "spark-submit", "#{my_app_jar}" ] },
 {"id" : "Activity2", "type" : "EmrActivity", "my_app_jar" : "s3://#{my_app_jar_bucket}/#{my_app2_name}.jar", "step" : [ "command-runner.jar", "spark-submit", "#{my_app_jar}" ] },
 {"id" :  "Activity3", "type" : "EmrActivity", "my_app_jar" : "s3://#{my_app_jar_bucket}/#{my_app3_name}.jar", "step" : [ "command-runner.jar", "spark-submit", "#{my_app_jar}" ] }
]
}
```

sethyates:


core/src/main/scala/com/krux/hyperion/DataPipelineDef.scala
```diff
@@ -1,15 +1,14 @@
 package com.krux.hyperion
 
 import scala.language.implicitConversions
-
-import com.amazonaws.services.datapipeline.model.{ ParameterObject => AwsParameterObject, PipelineObject => AwsPipelineObject }
+import com.amazonaws.services.datapipeline.model.{ParameterObject => AwsParameterObject, PipelineObject => AwsPipelineObject}
```
realstraw: We've been using mixed styles, no space after `{` and yes space after `{`. Which one do you think we should stick with?

sethyates: Personally I prefer no space. It also happens to be the IntelliJ default.

sethyates: But I'm happy to comply either way. This below was probably the result of IntelliJ auto reformatting, not something I intentionally did

realstraw: Let's use no space then. Probably easier for other people to make contributions since it's the default of IDE.

realstraw:
👍

PR #435, opened 2016-07-14
(Fixes #434) Adding ability for Parameters to be used for S3DistCpAct…

### LINKED ISSUE #434: S3DistCpActivity can't be passed Parameters
The example for S3DistCpActivity shows using a parameter for location, but you can't actually do that.  It's not allowed.  withSource and withDestination should both be able to accept HS3Uri and HHdfsUri objects.

hoangelos:
…ivity


sethyates:
LGTM

PR #440, opened 2016-07-14
(Fixes #439) Schedule.ondemand results in pipeline creation failure

### LINKED ISSUE #439: Schedule.ondemand results in pipeline creation failure
In 4.x, the Schedule is always serialised out.  This should not be serialised if Schedule is `ondemand`.

sethyates:


core/src/main/scala/com/krux/hyperion/DataPipelineDef.scala
```diff
@@ -36,13 +28,13 @@ object DataPipelineDef {
 
   implicit def dataPipelineDef2Json(pd: DataPipelineDef): JValue =
     ("objects" -> JArray(
-      (pd.defaultObject +: pd.schedule +: pd.workflow.toPipelineObjects.toList)
+      (List(pd.defaultObject) ++ pd.defaultObject.objects ++ pd.workflow.toPipelineObjects.toList)
```
realstraw: Does `pd.defaultObject +: pd.defaultObject.objects ++: pd.workflow.toPipelineObjects.toList` work?

sethyates: It does!  I'll fix in the squash.

core/src/main/scala/com/krux/hyperion/DataPipelineDef.scala
```diff
@@ -36,13 +28,13 @@ object DataPipelineDef {
...
         .map(_.serialize).sortBy(_.id).map(AdpJsonSerializer(_)))) ~
     ("parameters" -> JArray(
       pd.parameters.flatMap(_.serialize).map(o => AdpJsonSerializer(o)).toList))
 
   implicit def dataPipelineDef2Aws(pd: DataPipelineDef): Seq[AwsPipelineObject] =
-    (pd.defaultObject +: pd.schedule +: pd.workflow.toPipelineObjects.toList)
+    (List(pd.defaultObject) ++ pd.defaultObject.objects ++ pd.workflow.toPipelineObjects.toList)
```
realstraw: Unfortunately, this bit of code is also duplicated in https://github.com/krux/hyperion/blob/master/core/src/main/scala/com/krux/hyperion/DataPipelineDefGroup.scala#L66 you need to refactor or update that part as well.

sethyates:
Talk about falling down a rabbit hole.  This fix turned out to be heaps bigger than expected.

core/src/main/scala/com/krux/hyperion/DataPipelineDefGroup.scala
```diff
@@ -76,4 +55,35 @@ object DataPipelineDefGroup {
 
   final val DefaultNameKeySeparator = "#"
 
+  implicit class DataPipelineDefGroupOps(dpdg: DataPipelineDefGroup) {
```
realstraw: why do we need this implicit class instead of having those methods in `DataPipelineDefGroup`?

sethyates: The methods seem ancillary to the actual class itself, and seemed to be more "internal operations".  It wouldn't make sense for any subclasser to override them (or really call them TBH), so it seemed a way to clean up the class interface.

sethyates: I could even argue to add `private[hyperion]` to the implicit class.

realstraw:
👍

PR #444, opened 2016-08-03
(Fixes #443) Adding ability to accept multiple input and output data nodes in HiveActivity

### LINKED ISSUE #443: HiveActivity doesn't accept multiple input variables

aka11:
(Fixes #443) Adding ability to accept multiple input and output data nodes in HiveActivity


realstraw:
how would this work given multiple inputs and outputs? does it require the number of inputs and outputs to be the same?

aka11:
The number of inputs and outputs don't need to be same. They would be used in the hive script.

core/src/main/scala/com/krux/hyperion/activity/HiveActivity.scala
```diff
@@ -69,7 +69,7 @@ case class HiveActivity[A <: EmrCluster] private (
 
 object HiveActivity extends RunnableObject {
 
-  def apply[A <: EmrCluster](input: DataNode, output: DataNode, hiveScript: Script)(runsOn: Resource[A]): HiveActivity[A] =
+  def apply[A <: EmrCluster](input: Seq[DataNode], output: Seq[DataNode], hiveScript: Script)(runsOn: Resource[A]): HiveActivity[A] =
```
sethyates: This would be a breaking change for any scripts currently using this, so please provide an `apply` that takes `(input: DataNode, output: DataNode, hiveScript: Script)` and calls `apply(Seq(input), Seq(output), hiveScript)`.

core/src/main/scala/com/krux/hyperion/cli/EntryPoint.scala
```diff
@@ -6,6 +6,7 @@ import com.github.nscala_time.time.Imports._
 import com.krux.hyperion._
 import com.krux.hyperion.cli.Reads._
 import com.krux.hyperion.expression.Duration
+import org.json4s.BuildInfo
```
sethyates: Why was this added?

core/src/main/scala/com/krux/hyperion/activity/HiveActivity.scala
aka11: Ok, I will do that.

core/src/main/scala/com/krux/hyperion/cli/EntryPoint.scala
aka11: I was getting a compilation error without this.

realstraw: Must be your IDE setting. `import com.krux.hyperion._` should already import BuildInfo. This will only introduce conflicts and we don't need the version from `json4s`.

aka11: Ok I will check that and remove this import.

sethyates:
Please update `CHANGELOG.md`, `build.sbt` and `README.md` with a new version number.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 4.0.4 - 2016-08-04
```
sethyates: This is more a change than a fix, so I'd bump to `4.1.0`.

sethyates:
LGTM given it's 4.1.0, but I'll wait for @realstraw to give a +1 as well.

PR #451, opened 2016-08-08
(fixes #450) Update the default AMI from EBS to Instance Store

### LINKED ISSUE #450: Update the default AMI to use instance store AMIs instead of EBS backed

realstraw:


arijitdey:
:+1:

PR #453, opened 2016-08-10
(Fixes #448) fix a bug SparkActivity does not work on EMR release labels

### LINKED ISSUE #448: SparkActivity does not work with EMR release label 5.0.0
```
Error: Cannot load main class from JAR s3://myjar.jar with URI s3. Please specify a class through --class.
Run with --help for usage help or --verbose for debug output
```

the issue is the run-spark-step.sh does not use `--class` but spark-submit does. This probably affect release label 4.x.x as well

realstraw:
Note: although this changes the default behavior, this is not a breaking change because the default behavior is broken

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 4.3.1 - 2016-08-10
+### Fixed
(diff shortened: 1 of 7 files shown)


sethyates:
The committer who made the release label changes reported it worked for them, right?  If so, this would break them.

realstraw:
No, I've tested it does not work, because

1) SparkStep does not accept non s3 urls for jar location
2) Even if it does, it does not generate a compatible spark-submit arguments (where the class is in the wrong position)

sethyates:
OK, then the changes LSTM.

PR #454, opened 2016-08-11
(Fixes #447) Fix the conflict of ami and release label

### LINKED ISSUE #447: EMR cluster when using configuration to specify release label, the ami field is not overriden
It should shadow the AMI configuration if the release label field is present

realstraw:
- when both exists in configs

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/resource/EmrCluster.scala
+++ b/core/src/main/scala/com/krux/hyperion/resource/EmrCluster.scala
@@ -15,7 +15,7 @@ trait EmrCluster extends ResourceObject {
   def emrClusterFields: EmrClusterFields
   def updateEmrClusterFields(fields: EmrClusterFields): Self
 
-  def amiVersion = emrClusterFields.amiVersion
+  def amiVersion = if (releaseLabel.nonEmpty) None else emrClusterFields.amiVersion
   def withAmiVersion(version: HString): Self = updateEmrClusterFields(
     emrClusterFields.copy(amiVersion = Option(version), releaseLabel = None)
   )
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/krux/hyperion/resource/EmrCluster.scala
```diff
@@ -15,7 +15,7 @@ trait EmrCluster extends ResourceObject {
   def emrClusterFields: EmrClusterFields
   def updateEmrClusterFields(fields: EmrClusterFields): Self
 
-  def amiVersion = emrClusterFields.amiVersion
+  def amiVersion = if (releaseLabel.nonEmpty) None else emrClusterFields.amiVersion
```
sethyates: How is this happening to begin with?  Below, `withAmiVersion` resets `releaseLabel` to `None`.  Is it actually a bug in the `apply`?

realstraw: when config file have both ami and release label specified.

realstraw: yes, it's because of `apply` should we fix it there?

realstraw: actually, not `apply` but `defaultEmrClusterFields`

sethyates: Yeah I'd fix it in `defaultEmrClusterFields`.

realstraw: Done

jnewman:
LGTM

sethyates:
LGTM

PR #466, opened 2016-09-20
(Fixes #465) Updates AMI's

### LINKED ISSUE #465: Upgrade AMI's to 2017.09 HVM Instance Store
https://aws.amazon.com/amazon-linux-ami/

sethyates:


core/src/main/resources/reference.conf
```diff
@@ -122,18 +122,20 @@ hyperion {
       instance.type = "m1.small"
       instance.type = ${?HYPERION_EC2_INSTANCE_TYPE}
 
-      # Using Amazon Linux AMI (PV-IS) 2016.03.3 released on 2016-06-28
+      # Using Amazon Linux AMI (HVM-EBS) 2016.03.3 released on 2016-06-28
```
realstraw: Hour default instance type is `m1.small` so cannot use `HVM`, and data pipeline cannot easily use the storage provided by `EBS`, that's why we use instance storage by default. The reason we default to `m1.small` instead of the equivalent `t2` is that `t2` does not support AWS classic, and we do not want to break that function yet.

sethyates:
Any reason not to migrate to m3.medium? M1 is TWO generations behind

realstraw:
IIRC, t2.small is the one that's comparable to m1.small in terms of pricing and configuration, m3.medium might be a waste and more expensive. I'd love to drop AWS classic support though, since all AWS smaller instances are VPC only, but for now we need the classic instance support.

PR #471, opened 2016-10-19
Adds aws cp with profile activity

pkhetrapal:


contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,49 @@
+package com.krux.hyperion.activity
+
+import com.krux.hyperion.adt.{HS3Uri, HString}
+import com.krux.hyperion.common.{BaseFields, PipelineObjectId}
+import com.krux.hyperion.resource.{Ec2Resource, Resource}
+import com.krux.hyperion.expression.RunnableObject
+
+case class AwsS3CpActivity private(
+                                    baseFields: BaseFields,
```
realstraw: please make sure indentation is consistent with the rest of the codebase, and do not indent to `(`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,49 @@
...
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def useProfile(
+      profile: HString,
+      credentialsS3Uri: HS3Uri,
+      configS3Uri: HS3Uri) = updateShellCommandActivityFields(
+    shellCommandActivityFields.copy(script =
+      s"""`aws s3 cp $credentialsS3Uri ~/.aws/credentials`;
```
realstraw: is this going to fail if `~/.aws` does not exist?

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def apply(
+      sourceS3Path: HS3Uri,
+      destinationS3Path: HS3Uri
+    )(runsOn: Resource[Ec2Resource]): AwsS3CpActivity =
+
+    new AwsS3CpActivity(
+      baseFields = BaseFields(PipelineObjectId(AwsS3CpActivity.getClass)),
+      activityFields = ActivityFields(runsOn),
+      shellCommandActivityFields = ShellCommandActivityFields(s"aws s3 cp --recursive $sourceS3Path $destinationS3Path"),
```
realstraw: you might want to make `--recursive` an option, instead of making cp recusive

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
realstraw: Also, I think using `aws s3 sync` to sync over an `.aws` directory from s3 might be a better idea

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,62 @@
...
+  credentialsSource: Option[HString]
+) extends BaseShellCommandActivity {
+
+  type Self = AwsS3CpActivity
+
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def isRecursive() = copy(recursive = HBoolean.True)
```
sethyates: `isRecursive` would imply that this returns whether this is recursive or not.  Rather than setting it to be recursive.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,62 @@
...
+  def withProfile(profile: HString) = copy(profile = Option(profile))
+  def useCredentialsSource(credentialsSource: HString) = copy(credentialsSource = Option(credentialsSource))
+
+  def build() = updateShellCommandActivityFields(
```
sethyates: So do you envisage people calling `build()` in order to get a valid activity?  This isn't done like this anywhere else in this code base.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  type Self = AwsS3CpActivity
+
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def isRecursive() = copy(recursive = HBoolean.True)
+  def withProfile(profile: HString) = copy(profile = Option(profile))
+  def useCredentialsSource(credentialsSource: HString) = copy(credentialsSource = Option(credentialsSource))
```
sethyates: Rather than using this, you could take an Input and stage that input, and copy from there to `~/.aws`.  It might make a lot of this logic simpler if you did that.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,62 @@
...
+          aws s3 cp $$ARGS $sourceS3Path $destinationS3Path;
+        """
+    )
+  )
+}
+
+object AwsS3CpActivity extends RunnableObject {
+
+  def apply(
+      sourceS3Path: HS3Uri,
```
sethyates: Outdent one level.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,62 @@
...
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def isRecursive() = copy(recursive = HBoolean.True)
+  def withProfile(profile: HString) = copy(profile = Option(profile))
+  def useCredentialsSource(credentialsSource: HString) = copy(credentialsSource = Option(credentialsSource))
+
+  def build() = updateShellCommandActivityFields(
+    shellCommandActivityFields.copy(script =
+      s"""ARGS=""
+          if ${!credentialsSource.isEmpty} ; then
```
realstraw: it's better to use scala to construct the command rather than having too much logic in a bash script. It would read better and less prone to error

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -12,42 +12,46 @@ case class AwsS3CpActivity private(
...
-          if ${!credentialsSource.isEmpty} ; then
-            aws s3 sync ${credentialsSource.getOrElse("")} ~/.aws/;
-            ARGS="$$ARGS --profile ${profile.getOrElse("default")}";
-          fi
-          if $recursive ; then
-            ARGS="$$ARGS --recursive";
+  private var s3CpCommand: String = "aws s3 cp"
+
+  private def build() = {
+    if (recursive) {
```
sethyates: This is very procedural Java-esque code.  Please reformulate as more idiomatic Scala (e.g., without `var`).

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,65 @@
...
+  profile: Option[HString]
+) extends BaseShellCommandActivity with WithS3Input {
+
+  type Self = AwsS3CpActivity
+
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def makeRecursive() = copy(recursive = HBoolean.True).build()
```
realstraw: I suggest just call this method `recursive`. Also there is no need to call `build` here. You are just setting up the configurations for the script/command to be generated.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,65 @@
...
+  def withProfile(profile: HString) = copy(profile = Option(profile)).build()
```
realstraw: same here, no need to call `build`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  private var s3CpCommand: String = "aws s3 cp"
```
realstraw: please do not use `var` unless there is a very good reason

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  private def build() = {
```
realstraw: why not just override `def script`? and use configurations that you setup to build the command/script.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,101 @@
...
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def withRecursive() = copy(isRecursive = HBoolean.True)
+  def withOverwrite() = copy(isOverwrite = HBoolean.True)
+  def withProfile(profile: HString) = copy(profile = Option(profile))
+  def withAcl(cannedAcl: HString) = copy(cannedAcl = Option(cannedAcl))
+  def withExclude(pattern: HString) = copy(exclude = Option(pattern))
+  def withInclude(pattern: HString) = copy(include = Option(pattern))
+  def withSourceRegion(sourceRegion: HString) = copy(sourceRegion = Option(sourceRegion))
+  def withDestinationRegion(destinationRegion: HString) = copy(destinationRegion = Option(destinationRegion))
```
realstraw: are these all needed? If you need to have this level of freedom, I suggest just allow specifying strings as `cp` arguments instead of have a more restricted version. Or have one that allows user to add arguments to `cp` as strings and these ones call that method instead.

The reason why I'm suggesting this is that the main added value for this activity instead of a normal `ShellCommandActivity` is that is gets the profiles and put it to the right place. Maintain a map to all potential arguments to `aws s3 cp` seems to be a lot of effort, and you are not going all the way to map the potential argument and their types neither.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,101 @@
...
+    Seq(sourceS3Path, destinationS3Path)
+  ).flatten.mkString(" ")
+
+  val removeScript = if (isOverwrite) s"aws s3 rm --recursive $destinationS3Path;" else ""
+
+  override def script =
+    ShellCommandActivityFields(
+      s"""
+          if [ -n "$${INPUT1_STAGING_DIR}" ]; then
+            cp -R $${INPUT1_STAGING_DIR} ~/.aws;
```
realstraw: isn't this going to create a directory `$${INPUT1_STAGING_DIR}` in `~/.aws` as `~/.aws/$${INPUT1_STAGING_DIR}`?

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,101 @@
...
+    if (!include.isEmpty) Seq("--include", include.getOrElse("")) else Seq.empty[String],
+    if (!sourceRegion.isEmpty) Seq("--source-region", sourceRegion.getOrElse("")) else Seq.empty[String],
+    if (!destinationRegion.isEmpty) Seq("--region", destinationRegion.getOrElse("")) else Seq.empty[String],
+    Seq(sourceS3Path, destinationS3Path)
+  ).flatten.mkString(" ")
+
+  val removeScript = if (isOverwrite) s"aws s3 rm --recursive $destinationS3Path;" else ""
+
+  override def script =
+    ShellCommandActivityFields(
```
realstraw: there is no need to use this, just make `script` take the String

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
sethyates: I asked for these OOB as I have use cases for these. If we are creating objects the. I prefer to have them specify useful methods rather than having to just pass in a list of un type checked string arguments.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,101 @@
...
+          granteeTypeAndId.map { tuple => tuple.productIterator.mkString("=") }.mkString(",")
+        ).mkString("="):HString
+      )
+    )
+  }
+
+  val s3CpCommand  = Seq(
+    Seq("aws", "s3", "cp"),
+    if (isRecursive) Seq("--recursive") else Seq.empty[String],
+    if (!profile.isEmpty) Seq("--profile", profile.getOrElse("default")) else Seq.empty[String],
```
sethyates: Use `profile.nonEmpty` instead in this and all like it.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,101 @@
...
+    if (!cannedAcl.isEmpty) Seq("--acl", cannedAcl.getOrElse("bucket-owner-full-control")) else Seq.empty[String],
+    if (!grant.isEmpty) Seq("--grants", grant.getOrElse("")) else Seq.empty[String],
```
sethyates: You're already checking `isEmpty`. Why are you then providing a default to `getOrElse`? Instead you should be mapping the option like `grant.map(g => Seq("--grants", g))`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
realstraw: Talked to @pkhetrapal , we will change the implementation details to accommodate both your needs and have non-supported/new `aws cp` arguments easily supported without new releases as well.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+  def withSourceRegion(sourceRegion: HString) = addToAdditionArguments(Seq[HString]("--source-region", sourceRegion))
+  def withDestinationRegion(destinationRegion: HString) = addToAdditionArguments(Seq[HString]("--region", destinationRegion))
+  def withGrant(permission: HString, granteeTypeAndId: Seq[(String, String)]) = {
+    addToAdditionArguments(
+      Seq[HString]("--grants",
+        Seq(permission, granteeTypeAndId.map { tuple => tuple.productIterator.mkString("=") }.mkString(",")).mkString("=")
+      )
+    )
+  }
+  def withAdditionalArguments(arguments: Seq[HString]) = copy(additionalArguments = this.additionalArguments ++ arguments)
```
sethyates: Change this to `def withAdditionalArguments(arguments: HString*)`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+) extends BaseShellCommandActivity with WithS3Input {
+
+  type Self = AwsS3CpActivity
+
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def withOverwrite() = copy(isOverwrite = HBoolean.True)
+  def withRecursive() = addToAdditionArguments(Seq[HString]("--recursive"))
```
sethyates: Change this to `addToAdditionArguments("--recursive")`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+  def withProfile(profile: HString) =  addToAdditionArguments(Seq[HString]("--profile", profile))
```
sethyates: Change this to: `addToAdditionArguments("--profile", profile)` and fix spacing.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+  def withAcl(cannedAcl: HString) = addToAdditionArguments(Seq[HString]("--acl",cannedAcl))
+  def withExclude(pattern: HString) = addToAdditionArguments(Seq[HString]("--exclude", pattern))
+  def withInclude(pattern: HString) = addToAdditionArguments(Seq[HString]("--include", pattern))
+  def withSourceRegion(sourceRegion: HString) = addToAdditionArguments(Seq[HString]("--source-region", sourceRegion))
+  def withDestinationRegion(destinationRegion: HString) = addToAdditionArguments(Seq[HString]("--region", destinationRegion))
+  def withGrant(permission: HString, granteeTypeAndId: Seq[(String, String)]) = {
```
sethyates: Extraneous `{` and `}`.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+    addToAdditionArguments(
+      Seq[HString]("--grants",
+        Seq(permission, granteeTypeAndId.map { tuple => tuple.productIterator.mkString("=") }.mkString(",")).mkString("=")
```
sethyates: This doesn't seem to be the clearest way to do this.  Please re-evaluate it.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+      )
+    )
+  }
+  def withAdditionalArguments(arguments: Seq[HString]) = copy(additionalArguments = this.additionalArguments ++ arguments)
+
+  private def addToAdditionArguments(argument: Seq[HString]) = {
```
sethyates: Extraneous `{` and `}`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+    copy(additionalArguments = this.additionalArguments ++ argument)
+  }
+
+  private val removeScript = if (isOverwrite) s"aws s3 rm --recursive $destinationS3Path;" else ""
+
+  private val s3CpScript = "aws s3 cp %s %s %s;" format (additionalArguments.mkString(" "), sourceS3Path, destinationS3Path)
```
sethyates: Why not just use string interpolation instead of `format`?

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,79 @@
...
+
+  override def script = {
+    s"""
+          if [ -n "$${INPUT1_STAGING_DIR}" ]; then
+            cat $${INPUT1_STAGING_DIR}/credentials >> ~/.aws/credentials;
```
sethyates: If `~/.aws` doesn't exist, this will fail.  At the least you should `mkdir -p ~/.aws`.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 4.5.2 - 2016-10-20
+### Added
+- [#471] (https://github.com/krux/hyperion/pull/471) - Add support for `AWS S3 cp` CLI arguments in AwsS3CpActivity
```
sethyates: Remove the space between `] (`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,76 @@
...
+  type Self = AwsS3CpActivity
+
+  def updateBaseFields(fields: BaseFields) = copy(baseFields = fields)
+  def updateActivityFields(fields: ActivityFields[Ec2Resource]) = copy(activityFields = fields)
+  def updateShellCommandActivityFields(fields: ShellCommandActivityFields) = copy(shellCommandActivityFields = fields)
+
+  def withOverwrite() = copy(isOverwrite = HBoolean.True)
+  def withRecursive() = addToAdditionalArguments("--recursive")
+  def withProfile(profile: HString) = addToAdditionalArguments("--profile", profile)
+  def withAcl(cannedAcl: HString) = addToAdditionalArguments("--acl",cannedAcl)
```
sethyates: Space between `,` and `cannedAcl`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,76 @@
...
+  def withExclude(pattern: HString) = addToAdditionalArguments("--exclude", pattern)
+  def withInclude(pattern: HString) = addToAdditionalArguments("--include", pattern)
+  def withSourceRegion(sourceRegion: HString) = addToAdditionalArguments("--source-region", sourceRegion)
+  def withDestinationRegion(destinationRegion: HString) = addToAdditionalArguments("--region", destinationRegion)
+  def withGrant(permission: HString, granteeTypeAndId: Seq[(String, String)]) = addToAdditionalArguments("--grants",
+    granteeTypeAndId.map { case (grantType, id) => s"$grantType=$id" }.mkString(s"$permission=", ",", "")
+  )
+  def withAdditionalArguments(arguments: Seq[HString]) = copy(additionalArguments = this.additionalArguments ++ arguments)
+
+  private def addToAdditionalArguments(argument: HString*) = {
```
sethyates: Remove extraneous `{` and `}`.

sethyates: Actually, why have this when it is identical to `withAdditionalArguments`?

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,76 @@
...
+    copy(additionalArguments = this.additionalArguments ++ argument)
+  }
+
+  private val removeScript = if (isOverwrite) s"aws s3 rm --recursive $destinationS3Path;" else ""
+
+  private val s3CpScript = s"aws s3 cp ${additionalArguments.mkString(" ")} $sourceS3Path $destinationS3Path;"
+
+  override def script = s"""
+                          if [ -n "$${INPUT1_STAGING_DIR}" ]; then
+                            if [ ! -d ~/.aws ]; then
```
sethyates: You don't need to check for the existence first.  `-p` will not fail `mkdir` if it exists.

sethyates:
The commits need to be squashed when merging and the commit message of the squashed commit needs to follow the guidelines.

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -0,0 +1,73 @@
...
+  def withInclude(pattern: HString) = withAdditionalArguments(Seq("--include", pattern))
+  def withSourceRegion(sourceRegion: HString) = withAdditionalArguments(Seq("--source-region", sourceRegion))
+  def withDestinationRegion(destinationRegion: HString) = withAdditionalArguments(Seq("--region", destinationRegion))
+  def withGrant(permission: HString, granteeTypeAndId: Seq[(String, String)]) = withAdditionalArguments(
+    Seq(
+      "--grants",
+      granteeTypeAndId.map { case (grantType, id) => s"$grantType=$id" }.mkString(s"$permission=", ",", "")
+    )
+  )
+  def withAdditionalArguments(arguments: Seq[HString]) = copy(additionalArguments = this.additionalArguments ++ arguments)
```
realstraw: `def withAdditionalArguments(arguments: HString*)` is more consistent with the rest of the library. This will make the other methods that calls this look cleaner.

PR #474, opened 2016-11-01
(Fixes #353) Nicer handling for reducing iterable workflow expressions and empty workflows

### LINKED ISSUE #353: Enable .reduceLeft(_ ~> _) on a list of pipeline activities
Currently it does not work, as it expect to return `PipleineActivit[_ <: ResourceObject]` instead of `WorkExpression`.

At the moment, one either need to use `.foldLeft[WorkflowExpression](WorkflowNoActivityExpression)(_ ~> _)` or `.map(x => x: WorkflowExpression).reduceLeft(_ ~> _)`. Neither looks elegant.

Be able to simply use `.reduceLeft(_ ~> _)` would be ideal.

sethyates:
Current handling of an `Iterable` of `WorkflowExpression` or `PipelineActivity` fails on `reduceLeft` given empty `Iterable`.  Also, handling of `foldLeft` with a start of `WorkflowNoActivityExpression` is cumbersome.  This change makes this common use case cleaner.

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpressionImplicits.scala
+++ b/core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpressionImplicits.scala
@@ -8,13 +8,18 @@ import com.krux.hyperion.resource.ResourceObject
 trait WorkflowExpressionImplicits {
 
   implicit def workflowIterable2WorkflowExpression(activities: Iterable[WorkflowExpression]): WorkflowExpression =
-    activities.reduceLeft(_ + _)
+    activities.foldLeft(WorkflowExpression.empty)(_ + _)
 
   implicit def activityIterable2WorkflowExpression(activities: Iterable[PipelineActivity[_ <: ResourceObject]]): WorkflowExpression =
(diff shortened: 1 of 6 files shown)


core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpressionImplicits.scala
```diff
@@ -7,15 +7,21 @@ import com.krux.hyperion.resource.ResourceObject
 
 trait WorkflowExpressionImplicits {
 
+  def emptyWorkflow: WorkflowExpression = WorkflowNoActivityExpression
```
realstraw: I'd put this under the companion `object WorkflowExpression`, like how Scala collections `List.emptyList`

sethyates: OK well I've gone with `WorkflowExpression.empty` to mirror `Seq.empty` as opposed to `WorkflowExpression.emptyWorkflowExpression`.

PR #478, opened 2016-12-08
(Fixes #477) Make `initTimeout` as a optional global configuration

### LINKED ISSUE #477: Make `initTimeout` as a optional global configuration
When set, all resources will take this configuration and set it accordingly

arijitdey:
Adding ability to read the `initTimeout` from a global configuration file. It can also be set for a resource using`.withInitTimeout(5.hours)`.


core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
```diff
@@ -82,7 +82,8 @@ object Ec2Resource {
     region = Option(hc.ec2Region: HString),
     availabilityZone = hc.ec2AvailabilityZone.map(x => x: HString),
     subnetId = hc.ec2SubnetId.map(x => x: HString),
-    terminateAfter = hc.ec2TerminateAfter.map(x => x: HDuration)
+    terminateAfter = hc.ec2TerminateAfter.map(x => x: HDuration),
+    initTimeout = hc.emrInitTimeout.map(x => x: HDuration)
```
sethyates: Please don't use the emr setting here. You can have two different init timeouts in config one for ec2 and one for emr and they can both pull the same default.

README.md
```diff
@@ -29,7 +29,7 @@ Add Krux Hyperion as a dependency in your `build.sbt` or `Build.scala` as approp
 ```scala
 libraryDependencies ++= Seq(
   // Other dependencies ...
-  "com.krux" %% "hyperion" % "4.6.1"
+  "com.krux" %% "hyperion" % "4.6.2"
```
realstraw: Make this `4.7.0`

PR #498, opened 2017-05-31
(Fixes #497) Replace the Retry Implementation with Stubborn

### LINKED ISSUE #497: Replace the retry implementation with Stubborn
https://github.com/krux/stubborn

itsdawomb:
The retry implementation in Stubborn is more modular than the Retry code
previously included.

That said, we retain existing exp backoff+jitter behavior as a best
practice while allowing max_retries to still be configurable.


core/src/main/scala/com/krux/hyperion/client/AwsClientForDef.scala
```diff
@@ -67,6 +67,6 @@ case class AwsClientForDef(
    * Create and upload the pipeline definitions, if error occurs a full roll back is issued.
    */
   private def uploadPipelineObjects(): Option[AwsClientForId] =
-    UploadPipelineObjectsTrans(client, pipelineDef, maxRetry)().right.toOption
+    UploadPipelineObjectsTrans(client, pipelineDef)().right.toOption
```
realstraw: Why remove maxRetry is necessary? This breaks the design where `uploadPipelineObjects` should use the maxRetry defined here, which may not aways retrieve from the config.

itsdawomb: Yes, there is a conflict between the default maxRetry value configured in Stubborn and that which is configured in Hyperion. Is AwsClientForDef in Hyperion the primary candidate for external exposure (and hence increased customization via explicit reference to a config file)?

core/src/main/scala/com/krux/hyperion/client/AwsClient.scala
```diff
@@ -4,11 +4,12 @@ import com.amazonaws.auth.{DefaultAWSCredentialsProviderChain, STSAssumeRoleSess
...
 import com.amazonaws.services.datapipeline.DataPipelineClient
 import org.slf4j.LoggerFactory
-
 import com.krux.hyperion.DataPipelineDefGroup
+import com.krux.stubborn.Retryable
+import com.krux.stubborn.policy.ExponentialBackoffAndJitter
 
 
-trait AwsClient extends Retry {
+trait AwsClient extends Retryable with ExponentialBackoffAndJitter {
```
realstraw: You may want to override the `cap` and `base` to make sure, the behavior is the same as before.

itsdawomb: Good idea.

itsdawomb: OK so current behavior is picking 1000*(rand(0, n)) for n = 7, 9, and 13, where each n corresponds to one of the 3 default attempts. 

Stubborn behavior is a bit different -- the jitter is more refined:

override def retryDelay(attempt: Int): Int = Random.nextInt(super.retryDelay(attempt)),

where super.retryDelay takes values 6000, 12000, and 24000 by default. 

Given the change in how the randomness is injected, I think that the only question here is whether or not we should cap at a value less than 24000, given the proximity of 6k and 12k to 7k and 9k respectively, and general concerns of keeping the patchset free of magic numbers. 

What do you think?

itsdawomb: As discussed in person, we will adopt the new Stubborn cap and base override defaults with this patchset.

PR #516, opened 2017-11-26
(Fixes #507) Support STS token if environment variable is set

### LINKED ISSUE #507: Hyperion AWS client need to support STS tokens

realstraw:


core/src/main/scala/com/krux/hyperion/client/AwsClient.scala
```diff
@@ -1,12 +1,16 @@
 package com.krux.hyperion.client
 
-import com.amazonaws.auth.{DefaultAWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider}
+import com.amazonaws.auth.{
+  DefaultAWSCredentialsProviderChain, STSAssumeRoleSessionCredentialsProvider,
```
sushengloong: Nit: not in order

ssharma:
@realstraw  Any reason we're using `com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider` instead of `com.amazonaws.auth.EnvironmentVariableCredentialsProvider`. Assume role is ideally used for cross-account access or to refresh tokens in background thread. Our service users will not be MFA enabled and human users access should not be automatically refreshed.

realstraw:
@ssharma `com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider` is for supporting cross-account access

PR #528, opened 2018-02-16
(Fixes #527) Fix deep dependencies causing slow serialization

### LINKED ISSUE #527: Very deep dependencies will have causes slow serialization
When there is a long chain of dependencies, the `serialize` call in each activity object will be evaluated multiple times causing very slow execution time, and some times even stack overflow error.

realstraw:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 4.14.3 - 2018-02-09
+### Fixed
+- [#527](https://github.com/krux/hyperion/issues/527) - Updated all serialize calls on pipeline objects to lazy val to fixed an issue that very deep dependencies will have causes slow serialization
```
jpurnell01: nit: 'fix' -ed

jpurnell01: nit: 'will have causes' => 'will cause'

PR #542, opened 2018-05-21
Release Hyperion v5

realstraw:


contrib/activity/file/src/main/scala/com/krux/hyperion/contrib/activity/file/FileSplitter.scala
```diff
@@ -23,7 +22,13 @@ class FileSplitter(
 
     def isEmpty: Boolean = outputStreamWriter.isEmpty
 
-    def close(): Unit = outputStreamWriter.foreach(IOUtils.closeQuietly)
+    def close(): Unit = outputStreamWriter.foreach { o =>
+      try {
+        o.close()
+      } catch {
+        case e: Exception => e.printStackTrace()
```
prjasper: Add an explanatory message, so that if this occurs, it doesn't look like just a random stack trace in the log.

core/src/main/resources/reference.conf
```diff
@@ -144,14 +144,20 @@ hyperion {
...
       ami.version = "3.7"
       terminate = "8 hours"
       spark.version = "1.4.0.b"
+
+      # Only supported by LegacyEmrCluster
       env.uri = ${?HYPERION_EMR_ENV_URI}  # e.g. s3://bucket/org_env.sh
+
       inittimeout = ${?hyperion.resource.inittimeout}
+
+      release_label = "emr-5.12.0"
```
prjasper: Underscore looks strange in Scala.

core/src/main/scala/com/krux/hyperion/common/DurationConverters.scala
```diff
@@ -0,0 +1,26 @@
+package com.krux.hyperion.common
+
+import org.joda.time.Period
```
prjasper: Now Hyperion is running on the Java 8 SDK, this should be implemented with the standard java.time classes, not the org.joda.time. They're essentially the same (java.time is based on org.joda.time), so this should be a quick change.

.travis.yml
```diff
@@ -9,10 +9,10 @@ cache:
 language: scala
 
 scala:
-- 2.10.6
+- 2.12.4
```
jnewman: Waiting for travis to get .6?

core/src/test/scala/com/krux/hyperion/resource/SparkClusterSpec.scala
```diff
@@ -5,16 +5,11 @@ import com.typesafe.config.ConfigFactory
...
   implicit val hc = new HyperionContext(ConfigFactory.load("example"))
 
-  it should "handle release Label 4.x.x" in {
-    val cluster = SparkCluster().withReleaseLabel("emr-4.6.0")
-    cluster.standardBootstrapAction.length shouldBe 0
-    cluster.applications.length shouldBe 1
-  }
-
-  it should "be backwards compatible" in {
+  it should "be handle old spark versions" in {
```
jnewman: Nit - `be` is kinda weird

core/src/main/scala/com/krux/hyperion/common/DurationConverters.scala
realstraw: good point, will address this in future releases https://github.com/krux/hyperion/issues/543

core/src/main/resources/reference.conf
realstraw: yeah, but this is HOCON, and it's been like this for years, this line is just adding a default value, not introducing new configs, don't want to break things too much.

.travis.yml
realstraw: .4 was the latest version, when we did the PR to v5 branch, updated to `.6` now

core/src/test/scala/com/krux/hyperion/resource/SparkClusterSpec.scala
realstraw: Fixed

PR #544, opened 2018-05-25
Adds profile support in removeScript in AwsS3CpActivity

pkhetrapal:


contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -35,9 +36,21 @@ case class AwsS3CpActivity private(
   )
   def withAdditionalArguments(arguments: HString*) = copy(additionalArguments = this.additionalArguments ++ arguments)
 
-  private val removeScript = if (isOverwrite) s"aws s3 rm --recursive $destinationS3Path;" else ""
+  private val removeScript = isOverwrite match {
+    case HBoolean.True => profile match {
+      case Some(p) => s"aws s3 rm --recursive --profile $p $destinationS3Path;"
+      case _ => s"aws s3 rm --recursive $destinationS3Path;"
```
realstraw: let's use `None` instead of `_`

contrib/activity/definition/src/main/scala/com/krux/hyperion/activity/AwsS3CpActivity.scala
```diff
@@ -35,9 +36,21 @@ case class AwsS3CpActivity private(
...
+    }
+    case _ => ""
+  }
 
-  private val s3CpScript = s"aws s3 cp ${additionalArguments.mkString(" ")} $sourceS3Path $destinationS3Path;"
+  private val s3CpScript = {
+    profile match {
```
realstraw: There's no need to piggy back on additional arguments any more. You should simply build the command with `profile`

PR #548, opened 2018-07-06
(Fixes #547) Add silent failure to AwsS3CpAcitivyt

### LINKED ISSUE #547: Add silent failure option to the AWSS3CpActivity in Hyperion
We have noticed that the AWSS3CpActivity fails if the destination bucket is missing. As a result of which it cascade fails other activities. We need to add in an option to Hyperion's AWSS3cpActivity to fail silently even with non-zero exit code.

kyle-wright:


build.sbt
```diff
@@ -1,4 +1,4 @@
-val hyperionVersion = "5.1.0"
+val hyperionVersion = "5.2.0"
```
realstraw: let's make this `5.1.1` instead

kyle-wright: Changed to 5.1.1

realstraw:
I'll merge after travis check completes

PR #553, opened 2018-07-11
[552] : Adds support to execute custom script before running JarActivity

ssharma:
There is a need to increase process `ulimit` for a given session before executing `JarActivity`. This PR adds support to inject custom script before executing jar command


realstraw:
The use case seems like resource based configuration. Isn't it better to just build an AMI with ulimit pre-configured?

ssharma:
Yes, we can do that too. Just felt that would be more of a maintenance issue for a trivial ask

ssharma:
@realstraw We need to unblock soon. So, if this isn't something you're comfortable with, we'll go ahead and create custom AMI.

prashuj:
@realstraw  as you know it.. building and maintaining our own custom AMI's is a pain

realstraw:
@prashuj @prashuj It's no longer a pain to build and maintain AMIs, let's take the discussion offline from public git PR.

ssharma:
Not doing

PR #556, opened 2018-09-01
[5.1.3] Default to empty security groups when running in VPC

prjasper:


core/src/main/scala/com/krux/hyperion/resource/Ec2Resource.scala
```diff
@@ -70,8 +70,8 @@ object Ec2Resource {
     imageId = Option(hc.ec2ImageId: HString),
     runAsUser = None,
     associatePublicIpAddress = HBoolean.False,
-    securityGroups = hc.ec2SecurityGroup.toSeq,
-    securityGroupIds = hc.ec2SecurityGroupId.toSeq,
+    securityGroups = hc.ec2SubnetId.fold(hc.ec2SecurityGroup)(_ => None).toSeq,
```
realstraw: This only solves the issue when there subnet set in config, when the subnet is set using `.withSubnetId` it will still have conflict.

prjasper: Good point. See update. (Thanks for your help with it!)

Issue #560, opened 2018-10-15
Add support to run emr related activities on WokerGroup


qzhangsf:
**Status: Ready for Merge**

To support instance fleets in data pipelines, leverages `WorkerGroup` instead of `EmrCluster` to run emr related activities.

Changes needed:
1. Change `WorkerGroup` as a `trait` to be easily extended.
2. Keep consistency of `PipelineActivity` usage. This line https://github.com/krux/hyperion/blob/2529de63b3ea6bc5ce6ba924c19853d5f86c2f32/core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpression.scala#L10 converts workflow to `PipelineActivity[_]`. While the implicit conversion here https://github.com/krux/hyperion/blob/2529de63b3ea6bc5ce6ba924c19853d5f86c2f32/core/src/main/scala/com/krux/hyperion/workflow/WorkflowExpressionImplicits.scala#L16, requires `PipelineActivity[_ <: ResourceObject]` to convert an activity back to workflow expression. Reason to do this is to allow add customized `WorkerGroup` creation/termination activities to workflow.

PR for this issue: https://github.com/krux/hyperion/pull/561

Related PRs https://github.com/krux/krux-hyperion-extension/pull/84 and https://github.com/krux/krux-scala-data-processing/pull/1269

Issue #519, opened 2017-11-29
There are too much "magic" happening


realstraw:
there are a lot of implicit String to Some type and Some type to String conversions going on. It's probably better to reduce them.

PR #561, opened 2018-10-15
Add support to run emr related activities on WorkerGroup

qzhangsf:
Issue link: https://github.com/krux/hyperion/issues/560

Major changes:
1. Keep consistency of `PipelineActivity` usage: change all to `PipelineActivity[_ <: ResourceObject]`. 
2. Refactor `WorkerGroup`: change it to a `trait`.


core/src/main/scala/com/krux/hyperion/activity/PigActivity.scala
```diff
@@ -1,34 +1,34 @@
...
+import com.krux.hyperion.resource._
 
 /**
- * PigActivity provides native support for Pig scripts in AWS Data Pipeline without the requirement
- * to use ShellCommandActivity or EmrActivity. In addition, PigActivity supports data staging. When
- * the stage field is set to true, AWS Data Pipeline stages the input data as a schema in Pig
- * without additional code from the user.
- */
-case class PigActivity[A <: BaseEmrCluster] private (
+  * PigActivity provides native support for Pig scripts in AWS Data Pipeline without the requirement
```
realstraw: the previous doc style was correct

core/src/main/scala/com/krux/hyperion/activity/PigActivity.scala
```diff
@@ -1,34 +1,34 @@
...
-import com.krux.hyperion.resource.{Resource, BaseEmrCluster}
+import com.krux.hyperion.resource._
 
 /**
- * PigActivity provides native support for Pig scripts in AWS Data Pipeline without the requirement
- * to use ShellCommandActivity or EmrActivity. In addition, PigActivity supports data staging. When
- * the stage field is set to true, AWS Data Pipeline stages the input data as a schema in Pig
- * without additional code from the user.
- */
-case class PigActivity[A <: BaseEmrCluster] private (
```
realstraw: why remove the precise resource type?

core/src/main/scala/com/krux/hyperion/resource/EmrResource.scala
```diff
@@ -0,0 +1,13 @@
+package com.krux.hyperion.resource
+
+import scala.language.implicitConversions
+
+case class EmrResource(resource: Either[BaseEmrCluster, WorkerGroup])
+
+object EmrResource {
+
+  implicit def emrCluster2EmrResource(cluster: BaseEmrCluster): EmrResource = EmrResource(Left(cluster))
```
realstraw: let's also limit the use of those implicit conversion, we are actually trying to reduce those in the up coming v6 release. see: https://github.com/krux/hyperion/issues/519

core/src/main/scala/com/krux/hyperion/activity/PigActivity.scala
qzhangsf: Relax the resource type constraint to be able to run on either `EmrCluster` or `WorkerGroup`. 
The constructor `def apply[A <: BaseEmrCluster](script: Script)(runsOn: Resource[A]): PigActivity[A]` only accepts subtype of `BaseEmrCluster`, so I removed the resource type. The `EmrActivity` that replaces `MapReduceActivity` removes the resource type constraint, so I thought it is ok to apply similar thing to the other activities.

Any suggestion or alternates to relax the resource type constraint?

core/src/main/scala/com/krux/hyperion/resource/EmrResource.scala
qzhangsf: I introduced this `EmrResource` to relax the resource type constraints in the activities' constructors. In my first try, I used runtime instanceOf match to check resource type, but current solution should be better compared to runtime check. 

Any alternate to remove implicit conversion? could you please provide me some examples so that I can follow the same technique to revise the code?

core/src/main/scala/com/krux/hyperion/activity/PigActivity.scala
realstraw: This change drops support for older legacy emr cluster though, which is something we cannot do within one major release

qzhangsf: @realstraw 
The `LegacyEmrCluster` is also a subtype of `BaseEmrCluster`, which I think should work with the `EmrResource(resource: Either[BaseEmrCluster, WorkerGroup])`. 

Here is a comparison of the apply function of `PigActivity`:
old: `def apply[A <: BaseEmrCluster](script: Script)(runsOn: Resource[A]): PigActivity[A]`
new: `def apply(script: Script)(runsOn: EmrResource): PigActivity`, where `EmrResource` restricts the resource type.

I tried it with one pipeline using `LegacyEmrCluster`, e.g., https://github.com/krux/krux-scala-data-processing/blob/5498321479434f29a3634fafb3eb8d4da755d398/src/main/scala/com/krux/dataprocessing/pipeline/sambatv/SambaTvEventUserExtractor.scala#L18, and there is no error.

I am new to scala, not sure if any underlying magic makes it happen to be correct. Or just don't want to introduce more implicit conversion as you mentioned.

qzhangsf:
Hi @realstraw , have revised the code, and realized that current implementation already has support to run emr related activity on `WorkerGroup`. But still need some minor changes to allow customizing workflow. Also updated the Issue description https://github.com/krux/hyperion/issues/560.

realstraw:
also feel free to add yourself to https://github.com/krux/hyperion/blob/master/CONTRIBUTORS

ssharma:
@qzhangsf 🍪 🍸

Issue #568, opened 2018-12-24
Generalize Schedule / AdpSchedule to support expressions


mosche:
I'm currently evaluating hyperion to replace part of our AWS Data Pipeline tooling as hyperion is certainly more exhaustive. So far it looks great, there's just one major issue:

Currently `RecurringSchedule` (and also `AdpSchedule`) do not support the usage of expressions.
That however is really useful to delay the specification of an actual start time until activation of the pipeline. 

What's your thoughts on the change set below? Anything you could imagine?
https://github.com/krux/hyperion/compare/master...mosche:ExpressionSupportForSchedule
Happy to add some tests if worth investing...

In a next step `ActivateAction` could be extended to provide parameter values at activation time:
```
new ActivatePipelineRequest().withPipelineId(id).withParameterValues(....)
```

realstraw:
I didn't know that you can use expressions in schedule. Yeah, definitely something we'd like to add. I'll have a look at the diff this week. Thanks!

realstraw:
The diff looks good to me, let's add some tests and start a PR. Just curious of how can expression delay the start time until activation?

realstraw:
@mosche The diff looks good to me, let's add some tests and start a PR. Just curious of how can expression delay the start time until activation?

PR #569, opened 2018-12-31
Expression support for schedule

mosche:
@realstraw Here's the PR as discussed in https://github.com/krux/hyperion/issues/568
There isn't too much test coverage in general, so hard to say where to start. Do you have any preference?

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/Schedule.scala
+++ b/core/src/main/scala/com/krux/hyperion/Schedule.scala
@@ -1,11 +1,11 @@
 package com.krux.hyperion
 
 import com.github.nscala_time.time.Imports.{DateTime, DateTimeZone, Period}
-
+import com.krux.hyperion.Implicits._
+import com.krux.hyperion.adt.{HDateTime, HDuration, HInt}
 import com.krux.hyperion.aws.{AdpOnDemandSchedule, AdpRecurringSchedule, AdpRef}
 import com.krux.hyperion.common.{PipelineObject, PipelineObjectId, ScheduleObjectId}
 import com.krux.hyperion.expression.Duration
-import com.krux.hyperion.Implicits._
 
 /**
  * Schedule defines how a pipeline is run.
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/krux/hyperion/aws/AdpSchedule.scala
```diff
@@ -43,8 +41,8 @@ case class AdpRecurringSchedule(
   name: Option[String],
   period: String,
   startAt: Option[String],
-  startDateTime: Option[DateTime],
```
mosche: With this Json serialization in `AdpJsonSerializer` is obsolete as far as I can tell.
Any preference on whether to keep it or not?

core/src/main/scala/com/krux/hyperion/Schedule.scala
```diff
@@ -139,7 +140,7 @@ object Schedule {
         case None =>
           s
         case Some(dt) =>
-          s.copy(start = Option(dt.plus(by)))
+          s.copy(start = Option(dt.value.fold[HDateTime](_ plus by, _ + by)))
```
realstraw: Instead of using expressions for both cases, we should do this separately. If HDateTime is Joda DateTime, we should just do what we have before, and if it's an expression, we convert Period to Duration and use the generated expression in `DateTimeExp`

core/src/main/scala/com/krux/hyperion/expression/Expression.scala
```diff
@@ -140,6 +141,15 @@ trait BooleanExp extends TypedExpression with Evaluatable[Boolean]
 
 trait DateTimeExp extends TypedExpression {
 
+  def +(period: Period): DateTimeExp = Seq(
```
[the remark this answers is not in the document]
mosche: IMHO I don't think there is an alternative as long as `Period` is used in the `Schedule` interface. `Period` is significantly more powerful than a `Duration` and can only be expressed as a sum of `Durations`unless it is a `BaseSingleFieldPeriod`

core/src/main/scala/com/krux/hyperion/Schedule.scala
mosche: Not sure if I understand you correctly, but you might have missed that `plus` is Joda `DateTime#plus` and only the other case `+` is using an expression

realstraw: NVM, not sure what I meant neither. 🤦‍♂️

core/src/main/scala/com/krux/hyperion/expression/Expression.scala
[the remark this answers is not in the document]
realstraw: What I was suggesting is having the expression implementation free of other types. We can supply a converter from `Period` to `Duration` and make people do an explicit conversion when they want to use the expression.

It helps developers understand which part is evaluated by data pipeline and which part is evaluated at pipeline creation. It seems it's a common confusion, and separate expression with other more advanced data types helps a bit.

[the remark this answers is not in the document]
realstraw: hmm, I see what you mean. I think we are good on this.

mosche:
👍 I documented the change and bumped the version @realstraw

realstraw:
Thanks @mosche I'm merging this

PR #573, opened 2019-01-15
(Fixes #571) Use AWS defaults instead of hard code us-east-1

### LINKED ISSUE #571: Hyperion create pipeline undesirably ignores AWS_REGION environment variable

realstraw:


core/src/main/scala/com/krux/hyperion/client/AwsClient.scala
```diff
@@ -45,11 +43,11 @@ object AwsClient {
...
       }
 
-    DataPipelineClientBuilder
+    val defaultBuilder = DataPipelineClientBuilder
       .standard()
       .withCredentials(stsProvider.getOrElse(defaultProvider))
-      .withRegion(region)
-      .build()
+
+    region.map(defaultBuilder.withRegion(_)).getOrElse(defaultBuilder).build()
```
wandliu: Just sanity checking: So if regionId is None, then region is None, and we will use the defaultBuilder, which has the default configuration for region. Ok.

PR #587, opened 2019-09-13
Script run-python.sh now expects a requirements.txt file instead of requirements.pip

dllopis-sfdc:
For historical reasons peculiar to Krux, the `run-python.sh` script expected a `requirements.pip` file instead of a standard `requirements.txt` one. This change standardizes on `requirements.txt`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-val hyperionVersion = "5.5.1"
+val hyperionVersion = "5.6.0"
 val scala211Version = "2.11.12"
 val scala212Version = "2.12.9"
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -1,4 +1,4 @@
-val hyperionVersion = "5.5.0"
+val hyperionVersion = "5.5.1"
```
realstraw: This is a default behavior change, let's make it 5.6.0 instead.

dllopis-sfdc: That’s true—done.

Issue #441, opened 2016-07-14
SparkTaskActivity failing in 4.x.x


hoangelos:
Apparently, when using HadoopActivity, which the SparkTaskActivity uses, is complaining that command-runner.jar should be a URI from S3 or a local file.  It is a local file, so not sure if this is a bug in AWS or not.  But we could try to put the full path to command-runner.jar.  This wasn't surfaced before because the old version uses s3 location for this script.

Work around: Use SparkActivity with SparkStep(s) instead.

sethyates:
You mean `SparkTaskActivity`?

hoangelos:
Yes and I ironically broke it :-/

PR #592, opened 2019-10-25
Fix SparkTaskActivity to work with command-runner.jar

mosche:
We ran into a couple of issues when using `SparkTaskActivity` with the command-runner.

1st issue:  As described here: https://github.com/krux/hyperion/issues/441
Other than EmrStep, HadoopActivity must apparently provide the full path to the jar.

2nd issue: When using a jar in S3, the main class must be provided using the `--class` spark-submit option.
@realstraw I'm not entirely sure how to address this regarding your script runners. Any feedback there?

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
+++ b/core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
@@ -60,7 +60,11 @@ case class SparkTaskActivity private (
 
   override def objects = inputs ++ outputs ++ super.objects
 
-  private def sparkSettings: Seq[HString] = sparkOptions ++ sparkConfig.flatMap { case (k, v) => Seq[HString]("--conf", s"$k=$v") }
+  private def isCommandRunner = jarUri.toString.endsWith(EmrCommandRunner)
+
+  private def sparkSettings: Seq[HString] = sparkOptions ++
+    sparkConfig.flatMap { case (k, v) => Seq[HString]("--conf", s"$k=$v") } ++
+    sparkMainClass.filter(_ => isCommandRunner).toSeq.flatMap( main => Seq[HString]("--class", main.toString))
 
   lazy val serialize = new AdpHadoopActivity(
     id = id,
@@ -69,7 +73,7 @@ case class SparkTaskActivity private (
     mainClass = mainClass,
     argument = command.serialize +:
       sparkSettings.map(_.serialize) ++:
-      sparkJarUri.serialize +: sparkMainClass.map(_.toString) ++: arguments.map(_.serialize),
+      sparkJarUri.serialize +: sparkMainClass.filterNot(_ => isCommandRunner).map(_.toString) ++: arguments.map(_.serialize),
     hadoopQueue = hadoopQueue.map(_.serialize),
     preActivityTaskConfig = preActivityTaskConfig.map(_.ref),
     postActivityTaskConfig = postActivityTaskConfig.map(_.ref),
@@ -136,7 +140,7 @@ object SparkTaskActivity extends RunnableObject {
     baseFields = BaseFields(PipelineObjectId(SparkTaskActivity.getClass)),
     activityFields = ActivityFields(runsOn),
     emrTaskActivityFields = EmrTaskActivityFields(),
-    jarUri = EmrCommandRunner,
+    jarUri = EmrHadoopJarsDir.resolve(EmrCommandRunner).toString,
     sparkJarUri = jarUri,
     command = "spark-submit",
     sparkMainClass = None,
--- a/core/src/main/scala/com/krux/hyperion/activity/package.scala
+++ b/core/src/main/scala/com/krux/hyperion/activity/package.scala
@@ -1,5 +1,7 @@
 package com.krux.hyperion
 
+import java.net.URI
+
 import com.krux.hyperion.common.S3Uri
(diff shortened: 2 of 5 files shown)


realstraw:
Hmm, I remember I did some refactoring on how we were using the new `command-runner`, but have not paid much attention to `SparkTaskActivity`, I'll take a look

mosche:
@realstraw Did you get a chance to look into this?

realstraw:
@mosche not yet, hopefully, this weekend

realstraw:
This looks like a simple but valid approach. It works in your case, right? We are not using SparkTaskActivity that much on our end.

mosche:
Yes @realstraw 
We're currently using this workaround
```
val task: SparkTaskActivity
task
  .sparkMainClass.foldLeft(task)((t, m) => t.withSparkOption("--class", m.fullName)
  .copy(jarUri = "/var/lib/aws/emr/step-runner/hadoop-jars/command-runner.jar")
```

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
```diff
@@ -69,7 +71,7 @@ case class SparkTaskActivity private (
     mainClass = mainClass,
     argument = command.serialize +:
       sparkSettings.map(_.serialize) ++:
-      sparkJarUri.serialize +: sparkMainClass.map(_.toString) ++: arguments.map(_.serialize),
```
mosche: What I'm not sure about is if removing the main class here impacts the `scriptRunner` / `legacyScriptRunner` ...

core/src/main/scala/com/krux/hyperion/activity/package.scala
```diff
@@ -11,4 +13,5 @@ package object activity {
 
   final val EmrCommandRunner: String = "command-runner.jar"
 
+  final val EmrHadoopJarsDir: URI = URI.create("/var/lib/aws/emr/step-runner/hadoop-jars/")
```
realstraw: actually, this one does not make sense, it creates the folder locally, not on EMR

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
realstraw: ah, yes. So now I'm confused. The use case is to not use `run-spark-step-release-label.sh`? because it adds `--class` already https://github.com/krux/hyperion/blob/master/scripts/run-spark-step-release-label.sh#L53

core/src/main/scala/com/krux/hyperion/activity/package.scala
mosche: @realstraw Sorry, not sure how to read that comment ...
This folder exists on every EMR node (locally on the node) and contains the hadoop jars including the command-runner.jar.

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
mosche: As the title is saying, this is to fix SparkTaskActivity with the command-runner.jar. I haven't looked at run-spark-step-release-label.sh before, sorry I considered this to be some legacy thing up to now... 
But that's why I am asking how this change will impact the custom script runners before opening a PR.

Is there a reason you are using a custom script (that needs to be maintained/supported) if command-runner is provided for exactly that purpose? Is there any reason not to always use command-runner?

core/src/main/scala/com/krux/hyperion/activity/package.scala
mosche: Wait... are you talking about `URI.create`??? That creates a URI instance just like `new URI` not a folder or anything anywhere

realstraw: 🤦‍♂ ah, yes! of course!

core/src/main/scala/com/krux/hyperion/activity/SparkTaskActivity.scala
realstraw: I think the main reason is that so it downloads the jar from s3, IIRC command-runner.jar only runs local jars or in cluster-mode.

mosche: Ok 👍 Interesting though, why would you want to run something not in cluster mode?

mosche: So looks like the only feasible solution is to provide the main class differently depending on the runner type ...

realstraw: Yeah, I think this works

mosche: Great, i'll turn this into a PR. Let me know what version this should be, there's a bunch of other open PRs pending

realstraw: Let's make it 5.8.1

mosche: @realstraw sorry, somehow missed your reply... just made the final changes

realstraw: Thanks @mosche ! I'm merging it

Issue #341, opened 2016-02-19
WorkflowGraph simplification


realstraw:
The current workflow graph does not check the generated dependency graph is free of redundant dependencies (meaning `A->B->C` may generate a dependency with `A -> C`, which is not necessary)

It should by default (or have an option) to generate dependency graph free of redundant dependency.

related to: https://github.com/krux/hyperion/issues/190

mosche:
We've recently ran into Data Pipeline errors caused by the inclusion of transitive dependencies:

```
Errors: [May only specify up to '50' fields per object]
```

Problem is simply that this can quickly explode the number of fields per object.

PR #596, opened 2019-11-11
Correctly generate activity dependencies without any transitive ones

mosche:
This addressed issue https://github.com/krux/hyperion/issues/341. 

We've recently ran into Data Pipeline errors caused by the inclusion of transitive dependencies:
```
Errors: [May only specify up to '50' fields per object]
```
Problem is simply that this can quickly explode the number of fields per object.

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/workflow/WorkflowGraph.scala
+++ b/core/src/main/scala/com/krux/hyperion/workflow/WorkflowGraph.scala
@@ -1,125 +1,71 @@
 package com.krux.hyperion.workflow
 
-import scala.annotation.tailrec
-import scala.language.implicitConversions
-
 import com.krux.hyperion.activity.PipelineActivity
-import com.krux.hyperion.common.PipelineObjectId
 import com.krux.hyperion.resource.ResourceObject
 
 class WorkflowGraph private (
-  val flow: Map[PipelineObjectId, Set[PipelineObjectId]],
-  val activities: Map[PipelineObjectId, PipelineActivity[_ <: ResourceObject]]
+  val nodes: Seq[PipelineActivity[_ <: ResourceObject]],
+  val roots: Seq[PipelineActivity[_ <: ResourceObject]],
+  val leaves: Seq[PipelineActivity[_ <: ResourceObject]],
+  val dependencies: Seq[(PipelineActivity[_ <: ResourceObject], PipelineActivity[_ <: ResourceObject])]
 ) {
-
-  type Flow = Map[PipelineObjectId, Set[PipelineObjectId]]
-
-  def this() =
-    this(
-      Map.empty[PipelineObjectId, Set[PipelineObjectId]],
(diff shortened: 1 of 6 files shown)


mosche:
@realstraw and another once :)

core/src/main/scala/com/krux/hyperion/workflow/WorkflowGraph.scala
```diff
@@ -1,125 +1,64 @@
...
   /**
-   * Ids of PipelineObject that does not depend on anything
+   * The other graph follows this graph
    */
-  lazy val roots: Set[PipelineObjectId] =
-    flow.values.foldLeft(flow.keySet) { (rs, dependents) =>
-      rs -- dependents
-    } ++ isolates
+  def ~>(other: WorkflowGraph): WorkflowGraph =
+    new WorkflowGraph((nodes ++ other.nodes).distinct, other.leaves, (dependencies ++ other.dependencies).distinct)
```
realstraw: don't you need to add the dependencies that root nodes in `other` graph depending on leaves in `this` graph?

mosche: That's happening in the recursion here: https://github.com/krux/hyperion/pull/596/files#diff-db6229243a17f682aeff906c23aff577R17-R20

But agree, probably better to consolidate the logic in one place only

mosche: We are in fact building the `other` graph from the leaves of `this` graph. Initially I used this signature, but it's not really self explanatory ...
```def ~>(other: Seq[PipelineActivity[_ <: ResourceObject]] => WorkflowGraph): WorkflowGraph```

Thoughts?

File: core/src/main/scala/com/krux/hyperion/workflow/WorkflowGraph.scala [hunk 2]
```diff
@@ -1,125 +1,74 @@
...
-    )
-  }
+  type A = PipelineActivity[_ <: ResourceObject]
 
   /**
-   * isolates are both roots and leaves
+   * The other graph of the expression follows this graph
    */
-  lazy val isolates: Set[PipelineObjectId] = activities.keySet -- flow.keySet
+  def ~>(exp: WorkflowExpression): WorkflowGraph = {
```
realstraw: the other methods `~>` is operated on `WorkflowGraph`, it seems it's better to keep it consistent. I think I generally get what's going on in this PR, I would assume it would not be a lot of work to change this to `~> (other: WorkflowGraph)`?

mosche: Fair point 👍 I moved handling of dependencies out of the recursion into the graph. Certainly more consistent that way ...

realstraw:
LGTM, can you update the minor version and changelog?

mosche:
Done 👍

realstraw:
Thanks @mosche . I'm merging it

PR #597, opened 2019-11-20
Replace Joda time and nscala_time with Java time

damianxu88:
Risk: Low


realstraw:
@damianxu88 can you fix the build

core/src/main/scala/com/krux/hyperion/Schedule.scala
```diff
@@ -51,26 +52,28 @@ final case class RecurringSchedule private[hyperion] (
 
   def startDateTime(dt: HDateTime) = copy(start = Option(dt))
 
-
   def startThisHourAt(minuteOfHour: Int, secondOfMinute: Int) = {
-    val currentHour = DateTime.now.withZone(DateTimeZone.UTC).getHourOfDay
+    val currentHour = ZonedDateTime.now.withZoneSameLocal(ZoneOffset.UTC).getHour
```
realstraw: it should be `withZoneSameInstant` no?

core/src/main/scala/com/krux/hyperion/Schedule.scala
```diff
@@ -51,26 +52,28 @@ final case class RecurringSchedule private[hyperion] (
...
   def startThisMonthAt(dayOfMonth: Int, hourOfDay: Int, minuteOfHour: Int, secondOfMinute: Int) =
     startThisDayOfXAt(dayOfMonth, hourOfDay, minuteOfHour, secondOfMinute)(_.withDayOfMonth(_))
 
   private def startThisDayOfXAt(dayOfX: Int, hourOfDay: Int, minuteOfHour: Int,
-      secondOfMinute: Int)(dayOfFunc: (DateTime, Int) => DateTime) = {
-    val startDt = dayOfFunc(DateTime.now.withZone(DateTimeZone.UTC), dayOfX)
-      .withTime(hourOfDay, minuteOfHour, secondOfMinute, 0)
-    copy(start = Option(startDt: HDateTime))
+      secondOfMinute: Int)(dayOfFunc: (ZonedDateTime, Int) => ZonedDateTime) = {
+    val startDt = dayOfFunc(ZonedDateTime.now.withZoneSameLocal(ZoneOffset.UTC), dayOfX)
```
realstraw: same here, should be `sameInstant` I think

core/src/main/scala/com/krux/hyperion/Schedule.scala
```diff
@@ -51,26 +52,28 @@ final case class RecurringSchedule private[hyperion] (
...
+    val startDtTm = ZonedDateTime.of(
```
realstraw: you can use `startDt.with` instead, no?

core/src/main/scala/com/krux/hyperion/Schedule.scala
```diff
@@ -129,18 +132,18 @@ object Schedule {
 
   def onceAtActivation: RecurringSchedule = RecurringSchedule(end = Option(Left(1)), scheduleType = Cron)
 
-  def delay(schedule: Schedule, by: Duration, multiplier: Int): Schedule = {
+  def delay(schedule: Schedule, by: expression.Duration, multiplier: Int): Schedule = {
```
realstraw: you can import `com.krux.hyperion.expression.Duration` too

core/src/main/scala/com/krux/hyperion/Schedule.scala
```diff
@@ -129,18 +132,18 @@ object Schedule {
...
     import com.krux.hyperion.common.DurationConverters._
-    delay(schedule, by.asJodaPeriodMultiplied(multiplier))
+    delay(schedule, by.asPeriodDurationMultiplied(multiplier))
   }
 
-  def delay(schedule: Schedule, by: Period): Schedule = schedule match {
+  def delay(schedule: Schedule, by: PeriodDuration): Schedule = schedule match {
```
realstraw: let's change to use `java.time.Period` instead

core/src/main/scala/com/krux/hyperion/adt/HType.scala
```diff
@@ -131,12 +132,12 @@ object HBoolean {
...
 
-case class HDateTime(value: Either[DateTime, DateTimeExp]) extends HType {
+case class HDateTime(value: Either[ZonedDateTime, DateTimeExp]) extends HType {
 
-  val datetimeFormat = "yyyy-MM-dd'T'HH:mm:ss"
+  val datetimeFormat = DateTimeFormatter.ofPattern( "yyyy-MM-dd'T'HH:mm:ss")
 
   override lazy val serialize: String = value match {
-    case Left(dt) => dt.toDateTime(DateTimeZone.UTC).toString(datetimeFormat)
+    case Left(dt) => dt.withZoneSameLocal(ZoneOffset.UTC).format(datetimeFormat)
```
realstraw: same here. I think `.withZoneSameInstant` is the same as joda time's behavior, let me know if I'm not correct.

core/src/main/scala/com/krux/hyperion/common/DurationConverters.scala
```diff
@@ -1,26 +1,30 @@
...
 
   }
 
 }
+
+
+
+
+
+
```
realstraw: empty newlines

core/src/main/scala/com/krux/hyperion/Schedule.scala
damianxu88: just had a close look at `withZone` in Joda and it shifts the time field to the new time zone which is consistent with `withZoneSameInstant `, will change all where original was `withZone`

core/src/main/scala/com/krux/hyperion/Schedule.scala
damianxu88: as above, will change to `withZoneSameInstant `

core/src/main/scala/com/krux/hyperion/Schedule.scala
damianxu88: yes, I can use `startDt.with` with a LocalTime object passed in, will change

core/src/main/scala/com/krux/hyperion/Schedule.scala
damianxu88: oops, will do

core/src/main/scala/com/krux/hyperion/Schedule.scala
damianxu88: the problem with java.time.Period is it only covers year, month, week, day and java.time.Duration covers day, hour, minute, second, so the combination of the two is the equivalence of original org.joda.time.Period, that's why I created PeriodDuration to bridge the two

core/src/main/scala/com/krux/hyperion/adt/HType.scala
damianxu88: as above, will change

core/src/main/scala/com/krux/hyperion/common/DurationConverters.scala
damianxu88: will remove

core/src/main/scala/com/krux/hyperion/Schedule.scala
damianxu88: as a side note, `toDateTime(zone)` in joda time is the same as `withZoneSameLocal(zone)` with java time where date and time are preserved with only time zone applied

core/src/main/scala/com/krux/hyperion/Schedule.scala
realstraw: OK, so it seems using `java.time.Duration` is more appropriate here, no? Especially we are using it for arithmetic here.

damianxu88: Using `java.time.Duration` makes sense to enter this function, this does mean the `DateTimeExpr` where `+` arithmetic is carried out (and other associated places) will be changed to not have year, month, week fields

damianxu88:
@realstraw Can you have a look at the updated code?

core/src/main/scala/com/krux/hyperion/common/DurationConverters.scala
```diff
@@ -1,26 +1,23 @@
...
 
-
 /**
- * Handles the duration conversion between expression and Joda time
+ * Handles the duration conversion between expression and java.time.Duration
  */
 object DurationConverters {
 
-  implicit class AsJodaPeriod(duration: Duration) {
+  implicit class AsDuration(duration: Duration) {
```
realstraw: let's rename this to `AsJavaDuration`

damianxu88:
Ok, will do, just arrived in Sydney, waiting for the flight to Brisbane,
also looks like its still burning everywhere around Sydney, very smoky
here...

core/src/main/scala/com/krux/hyperion/common/DurationConverters.scala
damianxu88: Ok will do

damianxu88:
@realstraw Can you have another look?

realstraw:
LGTM, can you add an entry in CHANGELOG

damianxu88:
> LGTM, can you add an entry in CHANGELOG

Done, I've bumped up the version in the CHANGELOG to 5.7.3 following previous pattern but not sure if it is correct?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## 5.7.3 - 2019-12-03
```
realstraw: It should be 6.0.0-SNAPSHOT

damianxu88: Ok, updated now.

PR #598, opened 2020-01-03
Use json4s native backend instead of jackson

sushengloong:
This removes json4s' dependency on jackson from hyperion (aws libraries still use jackson) while using the same json4s API for serialization.

Upgrading sbt version is necessary to prevent `java.lang.OutOfMemoryError: Metaspace`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -7,7 +7,7 @@ val slf4jVersion    = "1.7.25"
 
 val nscalaTimeArtifact      = "com.github.nscala-time" %% "nscala-time"               % "2.18.0"
 val jodaConvertArtifact     = "org.joda"               %  "joda-convert"              % "2.0"    % "provided"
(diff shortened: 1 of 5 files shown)


sushengloong:
@realstraw this is ready for review.

realstraw:
It works to avoid jackson conflicts, but It probably wouldn't work for resolving json4s conflicts, as they share the same core.

sushengloong:
@realstraw The conflicts are with `com.fasterxml.jackson.**` and `org.json4s.jackson.**` specifically and not json4s core, right? I didn't see any shade rule for json4s core in other projects.

realstraw:
Hmm, IIRC that's because we keep json4s minor version consistent with spark. I remember there were some missing implicit def change we had to downgrade json4s version at some point

sushengloong:
@realstraw Got it. I will close the PR then. Thanks for your review.

PR #599, opened 2020-03-11
random subnet selection for ec2 resource

gcagrici:


core/src/main/scala/com/krux/hyperion/HyperionContext.scala
```diff
@@ -30,7 +32,7 @@ class HyperionContext(config: Config) {
   lazy val ec2Region = Try(config.getString("hyperion.aws.ec2.region")).toOption.getOrElse(region)
   lazy val ec2KeyPair = Try(config.getString("hyperion.aws.ec2.keypair")).toOption.orElse(keypair)
   lazy val ec2AvailabilityZone = Try(config.getString("hyperion.aws.ec2.availability_zone")).toOption
-  lazy val ec2SubnetId = Try(config.getString("hyperion.aws.ec2.subnet")).toOption
+  lazy val ec2RandomSubnetIds = Try(config.getStringList("hyperion.aws.ec2.random_subnets").asScala).toOption
```
realstraw: let's use `hyperion.aws.ec2.subnets` and maybe we can simply perform choosing the subnet here? Something like `lazy val ec2SubnetId = Try(pickOneRandom(config.getStringList("hyperion.aws.ec2.subnets"))).toOption`

gcagrici: 1. I will change the naming.
2. I did the random selection at the time of initializing the pipeline as the list of subnets can also be changed programmatically. `HyperionContext` is not the only place that allows the definition. Do you agree?

realstraw: I thought it would be a smaller/simpler change if everything happens at `HyperionContext`? It actually is the only place to interact with configurations.

core/src/main/scala/com/krux/hyperion/HyperionContext.scala
```diff
@@ -30,7 +32,8 @@ class HyperionContext(config: Config) {
   lazy val ec2Region = Try(config.getString("hyperion.aws.ec2.region")).toOption.getOrElse(region)
   lazy val ec2KeyPair = Try(config.getString("hyperion.aws.ec2.keypair")).toOption.orElse(keypair)
   lazy val ec2AvailabilityZone = Try(config.getString("hyperion.aws.ec2.availability_zone")).toOption
-  lazy val ec2SubnetId = Try(config.getString("hyperion.aws.ec2.subnet")).toOption
+  lazy val ec2SubnetId = Try(pickOneRandom(config.getStringList("hyperion.aws.ec2.subnets").asScala)).toOption
```
realstraw: +1 can you update https://github.com/krux/hyperion/blob/v6/CHANGELOG.md to include this change and stating that the config name change there too?

CHANGELOG.md
```diff
@@ -2,9 +2,10 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
-## 6.0.0-SNAPSHOT - 2019-12-03
+## 6.0.0-SNAPSHOT - 2020-03-13
 ### Changed
 - [#543](https://github.com/krux/hyperion/issues/543) - Replace joda time with java.time
+- [#577](https://github.com/krux/hyperion/issues/577) - Allow specifying multiple subnet IDs for EC2 resources in configuration (`hyperion.aws.ec2.subnets`) and randomly retrieve one upon pipeline initialization
```
realstraw: can you also add `hyperion.aws.ec2.subnet` is no longer supported?

PR #601, opened 2020-04-19
Capability to run Custom code when the pipeline is started

srikanthvissarapu:
Feature #W-7379481
Adding additional capability for pipelines to have custom behavior when pipelines are started.  Pipelines can optionally implement the PipelineLifeCycle trait and implement startpipeline API for custom functionality.

In our current scenario, we want to notify and track all the pipelines started via Hyperion for further monitoring & management.

### DIFF
--- a/core/src/main/scala/com/krux/hyperion/DataPipelineDefGroup.scala
+++ b/core/src/main/scala/com/krux/hyperion/DataPipelineDefGroup.scala
@@ -29,6 +29,14 @@ trait DataPipelineDefGroup
 
   def schedule: Schedule
 
+  def pipelineLifeCycle: PipelineLifeCycle = new PipelineLifeCycle {
+
+    override def startPipeline(id: String, name: String, status: String): Unit = {
+
+    }
+
+  }
+
   /**
    * No delay by default
    */
@@ -78,6 +86,7 @@ object DataPipelineDefGroup {
             dpdg.hc,
             dpdg.nameForKey(key),
             delayedSchedule(dpdg, idx),
+            dpdg.pipelineLifeCycle,
             () => workflow,
             dpdg.tags,
             dpdg.parameters
@@ -101,6 +110,7 @@ object DataPipelineDefGroup {
 
     def toAwsPipelineObjects: Map[WorkflowKey, Seq[AwsPipelineObject]] =
       objects.mapValues(_.map(_.serialize).toList.sortBy(_.id).map(AdpPipelineSerializer(_)))
+
 
     def toJson: JValue =
       ("objects" -> JArray(objects.values.flatten.map(_.serialize).toList.sortBy(_.id).map(AdpJsonSerializer(_)))) ~
--- a/core/src/main/scala/com/krux/hyperion/PipelineLifeCycle.scala
+++ b/core/src/main/scala/com/krux/hyperion/PipelineLifeCycle.scala
@@ -1,0 +1,7 @@
+package com.krux.hyperion
+
+trait PipelineLifeCycle {
+
+  def startPipeline(id: String, name: String, status: String)
+
+}
--- a/core/src/main/scala/com/krux/hyperion/DataPipelineDefGroupWrapper.scala
+++ b/core/src/main/scala/com/krux/hyperion/DataPipelineDefGroupWrapper.scala
@@ -4,22 +4,24 @@ import com.krux.hyperion.expression.{Duration, Parameter}
 import com.krux.hyperion.workflow.WorkflowExpression
 
 
-case class DataPipelineDefGroupWrapper private (
-  override val hc: HyperionContext,
-  override val pipelineName: String,
-  override val nameKeySeparator: String,
-  schedule: Schedule,
-  override val scheduleDelay: Option[Duration],
-  workflowsFunc: () => Map[WorkflowKey, WorkflowExpression],  // for delayed workfow execution
-  override val tags: Map[String, Option[String]],
-  override val parameters: Iterable[Parameter[_]]
-) extends DataPipelineDefGroup {
+case class DataPipelineDefGroupWrapper private(
+                                                override val hc: HyperionContext,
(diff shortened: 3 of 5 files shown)


realstraw:
In case of adding behavior on failures, you can simply register SNS in starport for listening on failures. For successful case, you can add activity that does that. In general, I'm not convinced adding more side effects during creation is necessary

rick-salesforce:
The point of adding a lifecycle callback in the Hyperion-based driver program is to let it define custom behavior in the callback.  

The implementation provided includes a default no-op so nobody is forced to provide additional code.

The initial use case motivating the lifecycle callback is to let the driver programs communicate their own payloads of information in the case of either failure to or successful submit a data pipeline.   Failure to submit/start a data pipeline can't be done in an activity in the pipeline.  Capturing the start could be done in the pipeline but it would be more & different code to communicate the payload from a pipeline activity rather than from the driver program.  By having the communication come from a callback in the driver, we can share the payload construction logic.

realstraw:
That's right, the point is we already have the supervisor program that knows success/failures where it sends out status to SNS. Programs can listen to those messages on SNS and react accordingly. Inserting arbitrary code on the driver where it has elevated permissions seems pretty dangerous and unnecessary.

rick-salesforce:
I'm not following 2 things.

1) The Hyperion-based driver is written by the developer and already contains custom logic to decide what to put into the pipeline.  Adding a callback to allow that driver to make its own notifications with its own payload of details seems comparable to the driver having access to config databases and other resources.  Maybe I don't understand the elevated permissions given to the driver.

2) It's great that there's a supervisor that handles success/failure across the board.  What I don't understand is how to get a custom payload (or extensions to the supervisor's payload) in notifications sent by the supervisor.  The plan in the lifecycle callback is to send an SNS notification so it would be consistent to subscribe to more SNS topics as long as the needed payload content can be included.

realstraw:
Point taken. So I'd vote for adding SNS to starport here https://github.com/krux/starport/blob/master/starport-core/src/main/scala/com/krux/starport/StartScheduledPipelines.scala#L242 current it sends out strings, but I'm happy to change it to a structured way in that case you would be able to get the name and maybe the AWS ID of the pipeline. And add notification on success cases too?

realstraw:
There's another project that current is yet to be open-sourced where all deployment happens in lambda and you'll be able to receive deployment result in SNS/SQS.

rick-salesforce:
I had a look a the suggested location StartScheduledPipelines.scala#L242  and then invested time to better understand the current responsibilities of Starport and the Pipeline class (and the  Hyperion logic that executes in the Pipeline class).  I've pretty much convinced myself that Starport can't do what we're trying to achieve in using the lifecycle hook proposed in this PR.  See point #2 in my prior comment.

To reiterate, Starport knows about success and failure on AWSDP create.  It doesn't know the 8-12 other things we plan to collect on AWSDP create.

Having written the above, I think I'd suggest to @srikanthvissarapu that we rename things in this PR.  Using the term "start" is technically inaccurate.  Perhaps the lifecycle stuff should be named to reflect that it's the PutPipeline API call (we call this upload in the Hyperion class) that is being exposed through the lifecycle callback?

@realstraw Your thoughts please. (Before Srikanth changes anything.)

realstraw:
The current implementation of the hook gives you only the name, pipelined, and state. What other 8-12 other things are you trying to collect?

rick-salesforce:
It's part of a new across-the-board addition to all of our workflows (not just Data Pipeline-based).  Some of the other attributes overlap with tags.  Not all of those use exactly the same set of values as the tags.  Others are things not expressed by tags.  For instance, "who" the data pipeline is working on behalf of (lists of 1 or more by type).  The pipelineId and state are what's needed from the context of the lifecycle event.  While it might be practical to push all into tags, that would defeat the uniformity of the new solution and add unnecessary complexity.

realstraw:
So gathering States, PipelineIds, and Names is enough for you?

rick-salesforce:
Yes.  As long as it's passed to code in the Pipeline class that's using Hyperion.  The other required info is not available anywhere else.

realstraw:
Ok, in that case, can we have starport do that instead?

core/src/main/scala/com/krux/hyperion/DataPipelineDefGroup.scala
```diff
@@ -29,6 +29,14 @@ trait DataPipelineDefGroup
 
   def schedule: Schedule
 
+  def pipelineLifeCycle: PipelineLifeCycle = new PipelineLifeCycle {
```
rkaruthethil, line 32: can we make this a singleton?

core/src/main/scala/com/krux/hyperion/PipelineLifeCycle.scala
```diff
@@ -0,0 +1,13 @@
+package com.krux.hyperion
+
+trait PipelineLifeCycle {
+
+  def onUploaded(id: String, name: String, status: Status.Value)
```
realstraw, line 5: let's add default noop implementation here

core/src/main/scala/com/krux/hyperion/PipelineLifeCycle.scala
```diff
@@ -0,0 +1,13 @@
...
+
+  def onCreated(id: String, name: String, status: Status.Value)
+
+}
+
+object Status extends Enumeration {
```
realstraw, line 11: Not a good idea to make this one on the top level module

core/src/main/scala/com/krux/hyperion/PipelineLifeCycle.scala
```diff
@@ -0,0 +1,13 @@
...
+  val SUCCESS, FAIL, SUCCESS_WITH_WARNINGS = Value
```
realstraw, line 12: Let's use `CamelCase` for those

PR #603, opened 2020-04-28
Capability to run Custom code when the pipeline is started - V6

srikanthvissarapu:


core/src/main/scala/com/krux/hyperion/DataPipelineDefGroupWrapper.scala
```diff
@@ -9,6 +9,7 @@ case class DataPipelineDefGroupWrapper private (
   override val pipelineName: String,
   override val nameKeySeparator: String,
   schedule: Schedule,
+  override val pipelineLifeCycle: PipelineLifeCycle,
```
realstraw, line 12: everything in the wrapper are evaluated expressions. Anyway we can make this outside of this?

core/src/main/scala/com/krux/hyperion/PipelineLifeCycle.scala
```diff
@@ -0,0 +1,13 @@
...
+
+trait PipelineLifeCycle {
+
+  def onCreated(id: String, name: String, status: Status.Value): Unit = {
+  }
+
+  def onUploaded(id: String, name: String, status: Status.Value): Unit = {
+  }
+
+}
```
realstraw, line 13: newline

core/src/main/scala/com/krux/hyperion/common/Status.scala
```diff
@@ -0,0 +1,5 @@
+package com.krux.hyperion.common
+
+object Status extends Enumeration {
```
realstraw, line 3: I mean since this is only applicable to `PipelineLifeCycle`, it makes more sense to put it within it's companion object

srikanthvissarapu: @realstraw I moved it to common, so it can reused (if needed) with in the framework. For example to capture any other DB, API response, etc.

realstraw:
```scala
object PipelineLifeCycle {
  object Status extends Enumeration {
    ...
  }
}
```

core/src/main/scala/com/krux/hyperion/DataPipelineDefGroupWrapper.scala
realstraw: Here's what I'm thinking, I don't think this belongs to the pipeline definition, as it's irrelevant in composing a pipeline. It seems a valid approach here is that we have those logging events https://github.com/krux/hyperion/blob/master/core/src/main/scala/com/krux/hyperion/client/UploadPipelineObjectsTrans.scala#L64 It would be good if we supply an implicit event handler and the default implementation is logging those, and you can add custom event handlers to the implicit scope and that will be used to inject whatever is needed.

realstraw: @srikanthvissarapu ^^

rick-salesforce: @realstraw Interesting point.  The lifecycle callbacks are not relevant to _composing_ a pipeline.  However, they are relevant to the driver knowing what's happening in UploadPipelineObjectsTrans.  I'm interpreting your suggestion as an alternative way to make them available to UploadPipelineObjectsTrans but I'm not following just where that is.

realstraw: Yes, that's what I'm trying to propose, there might be a way to inject actions to our AWS Client instead of copying that over to the wrapper. Let me sleep on this to see if I can crystalize what I'm thinking

realstraw: Actually, my original thought requires too much change, will probably do that in future refactoring. I'm good with this approach. Resolve this conversation

srikanthvissarapu:
@realstraw I updated the changelog. Please let me know if there is any other pending task from my side. Thanks! 
cc:@rick-salesforce

